Issue #33, opened 2020-01-21
Missing StartsWith(char8 c)


damianday:
We have 'public bool EndsWith(char8 c)' but not StartsWith(char8 c).
Oops should of mentioned in the String class :)

PR #34, opened 2020-01-22
Added String.StartsWith(char8)

rumbu13:
Correcting issue #33


bfiete:
Great, thanks. Now I suppose we are missing StartsWith/EndsWith with char32s - feel free to give those a shot if you care to!

PR #35, opened 2020-01-23
Added StartsWith(char32) & EndsWith(char32).

subramanivmk:
Added StartsWith(char32) & EndsWith(char32) similar to StartsWith(char8) & EndsWith(char8).

### DIFF
--- a/BeefLibs/corlib/src/String.bf
+++ b/BeefLibs/corlib/src/String.bf
@@ -1723,6 +1723,26 @@ namespace System
 			return Ptr[mLength - 1] == c;
 		}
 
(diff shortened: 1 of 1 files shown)


bfiete:
Nice. I modified it a bit:
In StartsWith, the GetEncodedLength is not necessary - the UTF8.Decode can take 'too large' of a size (ie: the entire buffer length).
In EndsWith, if the string buffer is shorter than the length of encodedLen then you end up with a negative ptr index. The new version just does a simple UTF8 backtrack and decodes and then compares from there.

Also I added a fast path for the (presumably) common 'low characters'.

PR #42, opened 2020-02-01
Fixed segfault in Environment.GetEnvironmentVariables on POSIX

lukpank:
The end condition on `endStr` was not checked causing segfault in `strlen()`.


bfiete:
Thanks!

Issue #40, opened 2020-01-29
[Enhancement] Array Reverse


damianday:
Reverse an array in-place.
Code modified from https://github.com/microsoft/referencesource/
[ArrayExt.zip](https://github.com/beefytech/Beef/files/4125556/ArrayExt.zip)

PR #90, opened 2020-03-20
Array reverse

damianday:
Fixes issue #40 .
Not sure if i should be using GetRef here for performance? will the compiler optimize this?

### DIFF
--- a/BeefLibs/corlib/src/Array.bf
+++ b/BeefLibs/corlib/src/Array.bf
@@ -147,6 +147,41 @@ namespace System
 			var sorter = Sorter<T, void>(&array.[Friend]mFirstElement, null, array.[Friend]mLength, comp);
 			sorter.[Friend]Sort(index, count);
 		}
+
+		// Reverses all elements of the given array. Following a call to this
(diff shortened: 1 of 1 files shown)


bfiete:
As far as optimization- in release mode, the bounds checks will optimize out even if array bounds checks were enabled (which usually they aren't for release mode). For debug mode you can manually omit the bounds checks either at the indexing: `arr[[Unchecked]i] = arr[[Unchecked]j];`
or you can add `[DisableChecks]` to the function declaration to just disable all checks in the method.

Interestingly, I compared how MSVC and Clang optimized with a std::vector<int64>. For a hand-rolled reverse equivalent to this one, the optimized C++ versions were quite a lot slower because they just didn't inline as well as BeefLang. In C++, however, std::reverse has a special-case for 8-byte elements which utilizes SIMD, which performs slightly better than BeefLang's Reverse on MSVC, but Clang's is almost twice as fast because it uses AVX whereas MSVC's only uses SSE.

It's in our roadmap to be able to do the same for those cases.

PR #93, opened 2020-03-24
Add String.Insert char32

damianday:
Use CalcNewSize in Insert.
Remove a duplicated Insert char8 by using default parameter.
Add new Insert method accepting char32.

### DIFF
--- a/BeefLibs/corlib/src/String.bf
+++ b/BeefLibs/corlib/src/String.bf
@@ -1432,14 +1432,30 @@ namespace System
 			mLength = newLength;
 		}
 
-		public void Insert(int_strsize idx, char8 c, int count = 1)
+		public void Insert(int_strsize idx, char8 c)
 		{
 			Contract.Requires(idx >= 0);
 
-			if (count == 0)
+			let newLength = mLength + 1;
+			if (newLength > AllocSize)
+				Realloc(CalcNewSize(newLength));
+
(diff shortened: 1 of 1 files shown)


damianday:
Good spot copy and paste failed me :/ I'll make some changes and test properly.

damianday:
@bfiete Should Append and Insert use UTF8.Encode? I can see its avoided I assume to avoid the overhead?

bfiete:
These char-appending methods are considered high priority for optimization, as they may be used to build large strings one character at a time. But- of course, for optimization the rule should be 'measure, don't assume'.

I timed some different implementations in Debug and Release. I found an even better implementation for Append. With a 2-byte UTF8 char, it's about 50% faster than versions that separately call GetEncodedLength in Release mode, with both the previous versions and a UTF8.Encode version being about the same speed as each other. In Debug, the previous version is about 20% faster than calling UTF8.Encode, but the new version at https://github.com/beefytech/Beef/commit/b5815927efd0e926a214dc5e2ecc4e686b16a491 is almost twice as fast.

And I do like having the separate versions of Append and Insert without the count parameter- it's by far the common case and it takes that extra beat of brainpower to wonder why there's that extra parameter you're not expecting in autocomplete, and there is a slight speed advantage when the method doesn't get inlined (ie: Debug mode).

bfiete:
Looks good.

PR #99, opened 2020-03-26
System.Threading.Parallel (experimental)

HydrogenC:
A PPL implemention of System.Threading.Parallel, untested, may have bugs. 
The only problem that cannot be solved is that I cannot get the pointer to every member of an IEnumerable. Otherwise, I can make Parallel.Foreach generic.

### DIFF
--- a/BeefRT/rt/Parallel.h
+++ b/BeefRT/rt/Parallel.h
@@ -1,0 +1,36 @@
+#pragma once
+
+#include <ppl.h>
+#include <vector>
+
+#include "BeefySysLib/Common.h"
+#include "BfObjects.h"
+
+namespace bf
+{
+	namespace System
+	{
+		namespace Threading
+		{
+			class Parallel : public Object
+			{
+			public:
+				BF_DECLARE_CLASS(Parallel, Object);
+
+				typedef void (*PInvokeFunc)();
+
+				BFRT_EXPORT static void InvokeInternal(void* funcs, int count);
+
+				typedef void (*PForFuncLong)(long long idx);
+				typedef void (*PForFuncInt)(int idx);
+
+				BFRT_EXPORT static void ForInternal(long long from, long long to, void* func);
+				BFRT_EXPORT static void ForInternal(int from, int to, void* func);
+
+				typedef void (*PForeachFunc)(void* item);
+
+				BFRT_EXPORT static void ForeachInternal(void* arrOfPointers, int count, void* func);
+			};
+		}
+	}
+}
--- a/BeefLibs/corlib/src/Threading/Parallel.bf
+++ b/BeefLibs/corlib/src/Threading/Parallel.bf
@@ -1,0 +1,55 @@
+using System;
+using System.Collections.Generic;
+
+namespace System.Threading {
+	public function void InvokeFunction();
+	public function void ForFunctionLong(int64 idx);
+	public function void ForFunctionInt(int32 idx);
+
+	public sealed class Parallel {
+		static extern void InvokeInternal(void* func1, int count);
+
+		public static void Invoke(InvokeFunction[] funcs)
+		{
+		    InvokeInternal(funcs.CArray(), funcs.Count);	
+		}
+
+		static extern void ForInternal(int64 from, int64 to, void* func);
+		static extern void ForInternal(int32 from, int32 to, void* func);
+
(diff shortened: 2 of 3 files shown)


BeefRT/rt/Parallel.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include <ppl.h>
+#include <vector>
```
HydrogenC: Redundant

bfiete:
You could have Foreach take a Span<T>, as both List<T> and T[] will implicitly convert.

HydrogenC:
Implemented with 6dcdf8d

HydrogenC:
The remaining problem now is that I can't refer to local variables in the function calls as only delegates allow doing that.

HydrogenC:
I tried to use delegate to replace function pointer, but it reported a link error on GetFuncPtr(): 
```
error LNK2019: public: static struct bf::System::ClassVData bf::CADNA::ForFunctionLong::sBfClassVData" (?sBfClassVData@ForFunctionLong@CADNA@bf@@2UClassVData@System@3@A)"public: static void __cdecl bf::CADNA::Program::Main(void)" (?Main@Program@CADNA@bf@@SAXXZ)
d:\ProgramData\CADNA\build\Debug_Win64\CADNA\CADNA.exe : fatal error LNK1120: 
```

bfiete:
Hm.. I'll see if I can repro that link error. At any rate, GetFuncPtr will fail at runtime if the delegate contains any state (ie: 'this' pointer, captures)...

HOWEVER- what I would do is something like this... (totally untested)

`
struct ForeachState<T>
{
  delegate void(T) mDelegate;
  Span<T> mArgs;

  public static void Call(Self* state)
  {
      for (let arg in state->mArgs)
         state->mDelegate(arg);
  }
}

static extern void ForeachInternal(void* func, void* state);

public static void Foreach<T>(Span<T> arr, delegate void(T item) dlg)
{
  ForeachState<T> state;
  state.mDelegate = dlg;
  state.mArgs = args;

  function void(ForeachState<T>* state) func = => ForeachState<T>.Call;
  ForeachInternal(func, &state);
}
`

And then set up the concurrency::parallel_for_each to call a lambda that calls back through the `func`, passing `state` back through.

HydrogenC:
Thanks a lot, I will try it as soon as possible.

bfiete:
Er- actually forget that. That doesn't actually work. The core idea of using a state struct to wrap up the delegate is the important part.

Your implementation of Parallel::ForeachInternal doesn't actually work, either. You must know the element size but you always assume the element size is sizeof(int).  Something that WOULD work would be to keep encapsulating the 'args' in the state object as I did, and then use concurrency::parallel_for and pass the index back through the State.Call method and then just index the mArgs when passing through to the mDelegate...

HydrogenC:
```
for (let arg in state->mArgs)
         state->mDelegate(arg);
```
Does beef actually support the "->" operator? I didn't find any thing about that in the docs.

bfiete:
My mistake, my brain switched to C for a second there. Should just be a dot operator.
```
for (let arg in state.mArgs)
         state.mDelegate(arg);
```

The dot operator works for pointers to structs, there's no arrow operator.

HydrogenC:
I have implemented C#-style ParallelState (originally called ParallelLoopState) and calling delegates, I will test it later.

HydrogenC:
Invoke and For works well, while Foreach causes access violation. I’ll fix it later.

HydrogenC:
The data structure of Beef Span<T> seems to be incompatible with a C-Array.

HydrogenC:
> Your implementation of Parallel::ForeachInternal doesn't actually work, either. You must know the element size but you always assume the element size is sizeof(int). 
Actually, my previous implementation does work because I actually pass a array of pointers to the elements instead of the elements themselves. 

> Something that WOULD work would be to keep encapsulating the 'args' in the state object as I did, and then use concurrency::parallel_for and pass the index back through the State.Call method. 
This sounds great and is feasible for Lists and Arrays, but I plan to support non-sequence containers such as Dictionaries and Deques. Thus, I think that passing array of pointers is the only way to implement that.

HydrogenC:
I Implemented Threading.Parallel for all platforms, but I would have to modify Dictionary<TKey, TValue> for a little bit.

PR #101, opened 2020-03-26
Stopwatch empty constructor

damianday:
Constructor should call Reset


bfiete:
Those values are already zeroed out - what does this fix?

damianday:
It only makes it clearer what the default state is, it's not really necessary if the default values match.

PR #114, opened 2020-04-09
Fix error tooltip position on multi-screen systems

### LINKED ISSUE #113: Error tooltip on second monitor causes DirectX crash
It happened a few time that my IDE crash without any apparent reason and with a DirectX-related error message.

Here's a screen capture of the last crash:
crash2

A quick Google search tells me that error 0x80070057 has the vague meaning "E_INVALIDARG"...

Unfortunately, I don't have a crash dump to give you and I haven't been able to reliably reproduce the bug.

Let me know if there's something I can do to provide you with more information next time this crash happens.

mandrelavoie:
Fixes #113 

SPI_GETWORKAREA retrieves the size of the work area on the primary display monitor only. On multi-screen setup, this was causing the desired position of the HoverWatch to be wrongly calculated (width was ending up negative).

### DIFF
--- a/BeefySysLib/platform/win/WinBFApp.cpp
+++ b/BeefySysLib/platform/win/WinBFApp.cpp
@@ -1131,7 +1131,19 @@ void WinBFApp::GetDesktopResolution(int& width, int& height)
 void WinBFApp::GetWorkspaceRect(int& x, int& y, int& width, int& height)
 {
 	RECT desktopRect;
-	::SystemParametersInfo(SPI_GETWORKAREA, NULL, &desktopRect, NULL);
+
+	if (::GetSystemMetrics(SM_CMONITORS) > 1) 
+	{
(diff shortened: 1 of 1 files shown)


bfiete:
Yeah, this is probably an okay fix, but the method is supposed to also keep windows from showing up only in the workspace (ie: subtracing windows regions such as the taskbar)... probably that code should just be refactored to be aware of which monitor it is referring to (passing in a desired virtual X/Y position, perhaps).

Also there's some window re-positioning code at https://github.com/beefytech/Beef/blob/master/BeefySysLib/platform/win/WinBFApp.cpp#L132 that IS aware of the desired location. There we probably need to find the right monitor through GetMonitorInfo and apply its work rect. 

Anyway- accepting this PR as-is since it's an improvement over what we had before.

PR #115, opened 2020-04-09
Changed mention of 'stack' to 'scope' in error msg

mandrelavoie:
I noticed that an error message was mentioning the use of the 'stack' keyword to allocate an object. Using it doesn't seem to work. I assume stack allocations used to be made with the 'stack' keyword, before it got changed to 'scope'?

If this is the case, this PR updates the error message.


bfiete:
Good catch. "stack" used to mean "scope::" before scope targets were added.

PR #133, opened 2020-04-25
Lossless Image Optimization

FlipFloop:


JamesOrson:
This only contains image files. Was there any code change?

FlipFloop:
No code change.

chances:
@FlipFloop Why is this necessary?

dsoronda:
i'll guess : reduced Download size ?

DerTee (approved):
I don't feel there's much benefit to scraping off a couple of bytes from image files, but it is a small improvement. Also, I've taken a look, there's no visual difference between the images before and after (if githubs image compare tools can be trusted), so there's no real downside ¯\_(ツ)_/¯

PR #153, opened 2020-04-28
Create ICollection

ChlorineX:
Created ICollection<T> and implemented it for List<T> and Dictionary<TKey, TValue>.


bfiete:
Looks good, thanks. Will probably change the CopyTo to operate on a Span<T> though.

ChlorineX:
By the way, are all types that implement IEnumerable<T> able to use in a foreach statement now?

PR #183, opened 2020-04-30
Fix build script

CompaqDisc:
Use absolute path rather than erroneously using relative path.


pmysl:
This PR doesn't solve the problem on macOS because `realpath` is not present by default

bfiete:
I'm working on this right this minute actually.

CompaqDisc:
This looks promising, https://github.com/whatwg/html-build/issues/90#issuecomment-235960102

PR #186, opened 2020-05-01
Various fixes and changes

HydrogenC:
1. Removed KeyValuePair<TKey, TValue> as it isn't actually needed
2. Use the builtin tuple of (TKey, TValue) to replace KeyValuePair
3. Removed a duplicated using
4. Renamed namespace System.Collections.Generic in the debug visualization files
5. The enumerator of a dictionary now enumerates (TKey, TValue*) instead of (TKey, TValue). (I wante to make it (TKey, ref TValue) at first, but found ref-types unusable in built-in tuples. Maybe make it possible to use ref-types in tuples in the future? )


HydrogenC:
Note: I've changed the return value of **Dictionary.Enumerator.Current** from (TKey, TValue) to **(TKey, TValue*)** _(this will make a ref-foreach over a dictionary possible)_. I haven't check every reference of Dictionary<TKey, TValue> so that this change may break some code, I will have a full check of BeefLibs later.

bfiete:
I was inspired by your work here to broaden how ref enumerators work. I just checked in some changes that build upon your work-

So now dictionary enumeration allows for both by-value and by-value enumeration:

```
for (var kv in dict)
    Console.WriteLine("{}={}", kv.key, kv.value);

for (var kv in ref dict)
    *kv.valueRef += 1000;
```

HydrogenC:
The ‘*kv.value’ should be reverted to ‘kv.value’ at PropertyBag.bf.

bfiete:
Already is: https://github.com/beefytech/Beef/blob/master/BeefLibs/corlib/src/PropertyBag.bf#L79

HydrogenC:
Sorry, I mistaken that.

PR #191, opened 2020-05-02
Base64 text encoding

Argentoz:
Added Base64 text encoding

### DIFF
--- a/BeefLibs/corlib/src/Text/Base64.bf
+++ b/BeefLibs/corlib/src/Text/Base64.bf
@@ -1,0 +1,75 @@
+namespace System.Text
+{
+	class Base64
+	{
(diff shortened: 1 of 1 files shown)


bfiete:
This looks pretty good, but- I'd expect the Encode to take a Span<uint8> for the incoming data. Also you're not reserving enough length in the string. You probably want a method to return the length of the encoded data. And you want the decoding to be able to take either a Span<uint8> for output or a List<uint8> - but the List<uint8> version should just reserve the right amount and then call to the Span version. Also you probably want to return a Result<void> to report failed decoding.

Argentoz:
> This looks pretty good, but- I'd expect the Encode to take a `Span<uint8>` for the incoming data. Also you're not reserving enough length in the string. You probably want a method to return the length of the encoded data. And you want the decoding to be able to take either a `Span<uint8>` for output or a `List<uint8>` - but the `List<uint8>` version should just reserve the right amount and then call to the Span version. Also you probably want to return a `Result<void>` to report failed decoding.

Can you give an example of how to use Span?

PR #200, opened 2020-05-03
Added String formatting constructor

EOG:
Added constructor to String that formats arguments based on pattern on creation based on existing code.
This allows to set formatted string values without introducing additional variable.

Solution collides with `[AllowAppend]public this(params String[] strs)` constructor, so formatting based on String values only is not possible. This is intentional.

### DIFF
--- a/BeefLibs/corlib/src/String.bf
+++ b/BeefLibs/corlib/src/String.bf
@@ -86,6 +86,12 @@ namespace System
 		    mAllocSizeAndFlags = (uint_strsize)bufferSize + (int_strsize)sizeof(char8*);
 		    mLength = 0;
 		}
+		
+		[AllowAppend]
+		public this(StringView format, params Object[] args) : this()
+		{
+			AppendF(format, params args);
+		}
 
 		[AllowAppend]
 		public this(String str)


bfiete:
I see why you may want something like this, but the only reason the String ctor already has so many overloads is because of append allocations: because the ctor arguments can be used to determine exactly how many characters of space need to be include in the original String allocation to accommodate the requested arguments. 

A formatted string cannot do this because you don't know how many characters you need until the last ToString is called on the arguments. Thus, even if you wanted this kind of ctor then you would still want overloads that allow you to specify how many characters to reserve. And then also you'd want versions that takes an IFormatProvider... 

Even so, I don't feel that string formatting could be considered a kind of "default behavior" for a constructor, and you have already pointed out the conflict with overload resolution on strings. This is one is the reasons format-style functions should only be used when there are very few other method overloads.

So I think it's really better to just stick with the explicit form of `scope String(reserveSize)..AppendF( ... )` (you did know about using the .. operator this way?)

bfiete:
That does bring up a good question - "what is the best output type for generic binary output"? I think `List<uint8>` is better than a String, for sure. But maybe an `ISequentialStream` interface is really the right kind of thing...

PR #206, opened 2020-05-04
A fix for #202

### LINKED ISSUE #202: The SDL2 Library is not working on linux
Does anyone know anything about this?

When I took the source code for the demo and tried compiling it on my linux machine, this error came up.

```
ERROR: Type could not be found (are you missing a using directive or library reference?) at line 5128:11 in /mnt/33CA263211FBF90F/Beef-master/BeefLibs/SDL2/src/SDL2.bf
   public Windows.HWnd window; // Refers to an HWND
          ^^^^^^^
ERROR: Type could not be found (are you missing a using directive or library reference?) at line 5129:11 in /mnt/33CA263211FBF90F/Beef-master/BeefLibs/SDL2/src/SDL2.bf
   public Windows.Handle hdc; // Refers to an HDC
          ^^^^^^^
Compile failed.
```

Is the problem in the demo or is the SDL library stuck to Windows?

ChlorineX:
Trying to fix #202.

### DIFF
--- a/BeefLibs/SDL2/src/SDL2.bf
+++ b/BeefLibs/SDL2/src/SDL2.bf
@@ -5121,7 +5121,6 @@ namespace SDL2
 			SDL_SYSWM_ANDROID
 		}
 
-
 		// FIXME: I wish these weren't public...
 		[CRepr]
 		public struct INTERNAL_windows_wminfo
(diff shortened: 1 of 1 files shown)


TheRawMeatball:
Did you test these patches? When I tested them, the compiler spat out this error instead:
```
Call parameter type does not match function signature!
i8* %0
 i32  %4 = call i32 @BfpSystem_SetCommandLine(i8* %0, i8* %1)
in function WinMain
LLVM ERROR: Broken function found, compilation aborted!
```
Besides, having to use a compiler preprocessor directive to filter out windows statements from a struct that has both internal _and_ windows in its name just feels wrong.

ChlorineX:
Sorry, I haven't checked all references.

PR #207, opened 2020-05-04
Base64 text encoding 3(Updated)

Argentoz:
Added Base64 text encoding
Example code:
```
	String InText = "SGVsbG8sIHdvcmxkIQ==";
	String OutText = scope String;
	Base64.DecodeString(InText, OutText);
	Console.WriteLine(OutText);
```
`Output: Hello, world!`

### DIFF
--- a/BeefLibs/corlib/src/Text/Base64.bf
+++ b/BeefLibs/corlib/src/Text/Base64.bf
@@ -4,7 +4,7 @@ namespace System.Text
 {
 	class Base64
 	{
-		static String lookup = new String("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/") ~ delete _;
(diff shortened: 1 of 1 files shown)


bfiete:
That does bring up a good question - "what is the best output type for generic binary output"? I think List<uint8> is better than a String, for sure. But maybe an ISequentialStream interface is really the right kind of thing...

bfiete:
But anyway- what I would have done here is had a DecodeData decode to a Span - because the output size can be trivially determined ahead of time.

Then have a DecodeString that reserves the correct amount of space in the string and then passes in the correct Span (rather than allocating once for List and once for String as you do now). Then also have a List<uint8> that also reserves the correct amount of space and then calls the DecodeData with the Span

Argentoz:
1

Argentoz:
close

PR #216, opened 2020-05-05
Add color support to System.Console

RogueMacro:
I added functionality for foreground and background colors that is cross-platform support. It uses `SetConsoleTextAttribute` for Windows and ansi escape sequences for Posix.


bfiete:
Nice!

RogueMacro:
Fixed gray and dark gray being switched

PR #277, opened 2020-05-17
add DeleteDictionaryAndItems function

kochol:


bfiete:
This made me realize I used the wrong name for the "DeleteDictionaryAndKeysAndItems" - it should have been "DeleteDictionaryAndKeysAndValues"

BeefLibs/corlib/src/System.bf
```diff
@@ -178,6 +178,18 @@ static
...
 	}
 
+	public static mixin DeleteDictionaryAndValues(var container)
+	{
+		if (container != null)
+		{
+			for (var value in container)
+			{
+				delete value.value;
+			}
```
JamesOrson, line 188: Wouldn't it be more appropriate to call the loop variable an `item`?

Ex:
```c#
for (var item in container)
{
    delete item.value;
}
```

bfiete: Or `element`. I often use `kv` for key/value pairs.

Issue #131, opened 2020-04-24
Can't build on macOS Catalina


pikatos:
bin/build.sh give me:

```
beef/Beef/IDEHelper/../extern/llvm-project_8_0_0/llvm/include/llvm/IR/Attributes.h:74:14: fatal error:
      'llvm/IR/Attributes.inc' file not found
    #include "llvm/IR/Attributes.inc"
             ^~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.```

**Comment by ByteSizeOtter** [comment]
Same issue here, but different error. 2019 MacBook Pro, Catalina (10.15.4). It initially failed at first but re-running it caused it to download LLVM and then started compiling. 2 hours later it failed, but with a different error from the other person. Ran it again just for kicks.

```
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/username/beeflang/Beef/jbuild
[  5%] Built target hunspell
[ 75%] Built target BeefySysLib
[ 93%] Built target IDEHelper
[ 94%] Built target BeefBoot
[100%] Built target BeefRT
Building BeefBuild_bootd
[******************************]
TIMING: Beef compiling: 16.2s
Linking BeefBuild_bootd...Undefined symbols for architecture x86_64:
  "_ffi_call", referenced from:
      bf::System::FFI::FFILIB::Call(bf::System::FFI::FFILIB::FFICIF*, void*, void*, void**) in libBeefRT_d.a(Internal.cpp.o)
  "_ffi_closure_alloc", referenced from:
      bf::System::FFI::FFILIB::ClosureAlloc(long, void**) in libBeefRT_d.a(Internal.cpp.o)
  "_ffi_prep_cif", referenced from:
      bf::System::FFI::FFILIB::PrepCif(bf::System::FFI::FFILIB::FFICIF*, bf::System::FFI::FFIABI, int, bf::System::FFI::FFIType*, bf::System::FFI::FFIType**) in libBeefRT_d.a(Internal.cpp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
FAIL: Exit code returned: 1

FAILED
```

nanjizal:
Beef/jbuild_d/Debug/bin/build/BeefBuild_bootd/vdata.o) was built for newer OSX version (10.14) than being linked (10.12)

bfiete:
[...]

Issue #120, opened 2020-04-15
Could not build on Macos 10.14 Mojave


phillvancejr:
cloned repository and ran bin/build.sh, but got the following error:
```
Beef/BeefySysLib/util/String.cpp:126:20: error: out-of-line definition
      of 'LastIndexOf' does not match any declaration in 'Beefy::StringView'
intptr StringView::LastIndexOf(char c, intptr startCheck) const

1 error generated.
make[2]: *** [BeefySysLib/CMakeFiles/BeefySysLib.dir/util/String.cpp.o] Error 1
make[1]: *** [BeefySysLib/CMakeFiles/BeefySysLib.dir/all] Error 2
```

bfiete:
There were a couple of Mac issues. Fixed in https://github.com/beefytech/Beef/commit/8709c92f2290bb0689bc0b229b3c28ce286507ff

phillvancejr:
clone the new repo and this time got a different error
```
Building BeefBuild_bootd
[******************************]
TIMING: Beef compiling: 56.6s
Linking BeefBuild_bootd...Undefined symbols for architecture x86_64:
  "_ffi_call", referenced from:
      bf::System::FFI::FFILIB::Call(bf::System::FFI::FFILIB::FFICIF*, void*, void*, void**) in libBeefRT_d.a(Internal.cpp.o)
  "_ffi_closure_alloc", referenced from:
      bf::System::FFI::FFILIB::ClosureAlloc(long, void**) in libBeefRT_d.a(Internal.cpp.o)
  "_ffi_prep_cif", referenced from:
      bf::System::FFI::FFILIB::PrepCif(bf::System::FFI::FFILIB::FFICIF*, bf::System::FFI::FFIABI, int, bf::System::FFI::FFIType*, bf::System::FFI::FFIType**) in libBeefRT_d.a(Internal.cpp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
FAIL: Exit code returned: 1

FAILED
```

pmysl:
It looks like a problem with libffi. I solved it by installing libffi from Homebrew and adding a path to lib in `[Configs.Debug.macOS]` and `[Configs.Release.macOS]` sections of `BeefBuild/BeefProj.toml`. You can also use libffi bundled with Beef. You'll find it here: `BeefySysLib/third_party/libffi/x86_64-apple-darwin19.3.0/.libs/libffi.a`

[...]

PR #286, opened 2020-05-21
Fix libffi issue on macOS

pmysl:
Fixes `libffi` issues mentioned in #120 and #131


bfiete:
What timing! I was staring at the link error on my mac right this second.. thanks!

PR #296, opened 2020-05-26
damianday

TheRealGammer:
This guy added some amazing and useful stuff, it would be cool to have them.


bfiete:
Yeah, he has- and he's submitted PRs. The only file change I actually see here is from https://github.com/beefytech/Beef/pull/101

PR #301, opened 2020-05-27
Array Copy

damianday:
Added copying without specifying offsets.


bfiete:
Hm- I think we should be able to strengthen the generic constraints on these methods to `where T2 : operator explicit T` now...

PR #302, opened 2020-05-27
Array CopyTo for spans

damianday:
CopyTo, allows copying memory from an array to a span.

### DIFF
--- a/BeefLibs/corlib/src/Array.bf
+++ b/BeefLibs/corlib/src/Array.bf
@@ -274,12 +274,12 @@ namespace System
 			Internal.MemCpy(destination.Ptr, &GetRef(0), strideof(T) * mLength, alignof(T));
 		}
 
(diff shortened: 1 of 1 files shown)


bfiete:
I don't think there should be a `length` on the second one. It's trivial to create a new span with a shorter length if you don't want to fill out the whole span...

damianday:
Thats true, I've removed the length parameter.

bfiete:
Just one more thing- the length check doesn't take into account the offset.
`Debug.Assert(destination.Length >= mLength);`
should be
`Debug.Assert((uint)srcOffset + (uint)destination.[Friend]mLength <= (uint)mLength);`
The `[Friend]mLength` part is sortof optional because it's an `[Inline]` anyway, but it's still slightly faster in debug mode (especially on the LLVM backend).

bfiete:
Oh, actually you set this up the inverse of what I expected- the length copied should be the size of the span, not the size of the array.

PR #306, opened 2020-05-27
Stack

damianday:
Tried to keep the implementation as close to Queue/List as possible.
I've included tests to ensure everything works as expected - which are commented out.

### DIFF
--- a/BeefLibs/corlib/src/Collections/Stack.bf
+++ b/BeefLibs/corlib/src/Collections/Stack.bf
@@ -1,0 +1,485 @@
+// This file contains portions of code released by Microsoft under the MIT license as part
+// of an open-sourcing initiative in 2014 of the C# core libraries.
+// The original source was submitted to https://github.com/Microsoft/referencesource
+
+namespace System.Collections
+{
+	using System;
+	using System.Diagnostics;
+	using System.Threading;
+
+	/// A simple stack of objects.  Internally it is implemented as an array,
+	/// so Push can be O(n).  Pop is O(1).
+	public class Stack<T> : IEnumerable<T>//System.Collections.ICollection, IReadOnlyCollection<T>
+	{
+		const String InvalidOperation_EnumNotStarted = "Enumeration not started";
+		const String InvalidOperation_EnumEnded = "Enumeration ended";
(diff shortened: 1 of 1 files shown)


bfiete:
Just wondering - what is the use case for an explicit stack vs a List? It just removes random-access functionality, but a list already has Add, PopBack, and the Back property...

damianday:
Yeah I've seen a few stack implementations which just wrap a an array list. The only gain I can think of is familiarity, especially with people coming from c# and a cleaner API.

JamesOrson:
I think the benefit of a Stack here would depend on the implementation of the List people generally use as a Stack.

A Stack can be made to be incredibly efficient using a LinkedList. Insertion and deletion are still O(1) operations that way. Depending on the implementation of the List, the List may have more memory allocated than necessary for the size of the Stack.

So say the List has allocated 8 spots for items and all of them are filled up. If I go to add an item to the "Stack", in a normal implementation of an expandable collection, the List's capacity would be doubled to 16 spots leaving the List with 7 empty spots that may never be used.

A Stack implemented with a LinkedList would only consume as much memory as is being used at any given time. Unless the Stack is still slab allocated the same way as the List, with nodes on standby.

JamesOrson:
> I think the benefit of a Stack here would depend on the implementation of the List people generally use as a Stack.
> 
> A Stack can be made to be incredibly efficient using a LinkedList. Insertion and deletion are still O(1) operations that way. Depending on the implementation of the List, the List may have more memory allocated than necessary for the size of the Stack.
> 
> So say the List has allocated 8 spots for items and all of them are filled up. If I go to add an item to the "Stack", in a normal implementation of an expandable collection, the List's capacity would be doubled to 16 spots leaving the List with 7 empty spots that may never be used.
> 
> A Stack implemented with a LinkedList would only consume as much memory as is being used at any given time. Unless the Stack is still slab allocated the same way as the List, with nodes on standby.

I see that this implementation still follows the methodology of "When I am full, double my capacity", so there really is not a difference in this implementation from just using a List with `Add`, `PopBack`, and `Back`.

damianday:
I'll close the pull for now, because their is not much benefit over the list implementation.

PR #386, opened 2020-07-01
Change File->Preferences->Settings to File->Settings

wyattscarpenter:
I should note that I have not tried to compile this code, I am hoping the change is simple enough to not be wrong.

### DIFF
--- a/IDE/src/IDEApp.bf
+++ b/IDE/src/IDEApp.bf
@@ -5006,9 +5006,7 @@ namespace IDE
             AddMenuItem(subMenu, "&Save File","Save File", new => UpdateMenuItem_HasActiveDocument);
             AddMenuItem(subMenu, "Save &As...", "Save As", new => UpdateMenuItem_HasActiveDocument);
 			AddMenuItem(subMenu, "Save A&ll", "Save All");
-			let prefMenu = subMenu.AddMenuItem("&Preferences");
-			//prefMenu.AddMenuItem("&Keyboard Shortcuts", null, new (evt) => { ShowKeyboardShortcuts(); });
-			AddMenuItem(prefMenu, "&Settings", "Settings");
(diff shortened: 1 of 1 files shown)


IDE/src/IDEApp.bf
```diff
@@ -5006,9 +5006,7 @@ namespace IDE
             AddMenuItem(subMenu, "&Save File","Save File", new => UpdateMenuItem_HasActiveDocument);
             AddMenuItem(subMenu, "Save &As...", "Save As", new => UpdateMenuItem_HasActiveDocument);
 			AddMenuItem(subMenu, "Save A&ll", "Save All");
-			let prefMenu = subMenu.AddMenuItem("&Preferences");
-			//prefMenu.AddMenuItem("&Keyboard Shortcuts", null, new (evt) => { ShowKeyboardShortcuts(); });
-			AddMenuItem(prefMenu, "&Settings", "Settings");
+			AddMenuItem(subMenu, "S&ettings", "Settings");
```
JamesOrson, line 5009: It was `"&Settings"` before. Is `"S&ettings"` intentional?

wyattscarpenter: Yes, otherwise it would conflict with the s in save, right?

JamesOrson: I am not sure how this stuff works. Was more curious as to the reasoning why the & was moved.

wyattscarpenter: I am also not sure how this works, but the impression I got was that the & controls the letter that gets to be the hotkey, so I had to make "&Settings" into "S&ettings" so it wouldn't conflict with "&Save File" now that they are on the same level of the menu.

wyattscarpenter:
There are more items in the prefMenu now so this change no longer makes sense.

PR #394, opened 2020-07-10
Remove unnecessary delete statements

JamesOrson:
If a value is null, the C++ delete statement will do nothing, so there is bo need to explicitly check if the value is null.

### DIFF
--- a/BeefySysLib/FileHandleStream.cpp
+++ b/BeefySysLib/FileHandleStream.cpp
@@ -13,8 +13,7 @@ FileHandleStream::FileHandleStream()
 
 FileHandleStream::~FileHandleStream()
 {
-	if (mCacheBuffer != NULL)
-		delete mCacheBuffer;
+	delete mCacheBuffer;
 	if (mFileHandle != NULL)	
 		::CloseHandle_File(mFileHandle);	
 }
--- a/BeefySysLib/FileStream.cpp
+++ b/BeefySysLib/FileStream.cpp
@@ -16,8 +16,7 @@ FileStream::FileStream()
 
 FileStream::~FileStream()
 {
-	if (mCacheBuffer != NULL)
-		delete mCacheBuffer;
+	delete mCacheBuffer;
 	if (mFP != NULL)
 		fclose(mFP);
 }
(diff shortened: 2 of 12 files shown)


IDEHelper/Compiler/BfIRBuilder.cpp
```diff
@@ -1166,11 +1166,7 @@ bool BfIRBuilder::HasExports()
 
 BfIRBuilder::~BfIRBuilder()
 {
-	if (mIRCodeGen != NULL)
-	{
-		mIRCodeGen->mStream = NULL;
```
JamesOrson, line 1171: This setting to NULL seemed unnecessary, as this destructor is cleaning up the `mIRCodeGen` pointer entirely and this is a member variable of the class.

IDEHelper/Compiler/BfIRBuilder.cpp
```diff
@@ -1572,12 +1568,8 @@ void BfIRBuilder::SetBackend(bool isBeefBackend)
 
 void BfIRBuilder::RemoveIRCodeGen()
 {
-	if (mIRCodeGen != NULL)
-	{
-		mIRCodeGen->mStream = NULL;
```
JamesOrson, line 1577: This setting to NULL seemed unnecessary, as this function is cleaning up the `mIRCodeGen` pointer entirely and this is a member variable of the class.

IDEHelper/Compiler/BfIRBuilder.cpp
bfiete: This `mIRCodeGen->mStream = NULL` was not a no-op. It was to keep the derived classes `BeIRCodeGen` and `BfIRCodeGen` from attempting to delete `mStream` in their destructors since normally they own a stream but in this particular case it was set to read from a stream owned by `BfIRBuilder`.

JamesOrson: Gotcha. Was trying to think of a case where this would be needed. Thanks for the clarification.

elfring:
Thanks for your small source code improvement.

PR #428, opened 2020-07-27
Update Socket.bf

damianday:
Fixed select polling problem.
Implemented SetBlocking.

### DIFF
--- a/BeefLibs/corlib/src/Net/Socket.bf
+++ b/BeefLibs/corlib/src/Net/Socket.bf
@@ -223,7 +223,7 @@ namespace System.Net
 
 			mHandle = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
 			
-			SetBlocking();
+			SetNonBlocking(true);
 
 			if (mHandle == INVALID_SOCKET)
 			{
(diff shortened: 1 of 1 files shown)


damianday:
Just on a side note I've not made the set blocking public because it could really use a get blocking method and then make it a simple property

bfiete:
What is this change in Select? That seems wrong... are there not 1000000 usecs in a second?

Also what was the point of pulling out the blocking code into SetBlocking, if it's private anyway? Also it's kindof misnamed because it sets the socket to non-blocking.

damianday:
Bugger your right, i was passing milliseconds no wonder it wasn't right 🤔

I figure the blocking code will eventually have methods for getting and setting blocking state, so just did a some refactor, and.yeah name is a bit misleading, c# calls ot InternalSetBlocking

bfiete:
Normally I use milliseconds, but the underlying `select` uses microseconds and I consider this class to be a "thin wrapper" so I wanted to stick as closely as possible to the underlying API.

Now I do kind of feel that might be wrong since it's inconsistent with just about every other int-based time span in corlib.  Maybe there should be another `Select` that can accept a TimeSpan that can have microsecond accuracy... or a `SelectMicro` or something...

damianday:
I've reverted the select behaviour and modified the blocking code.

I don't know if its worth the hassle of having 2 select methods, I think most people would prefer to use more performant APIs instead when available.

bfiete:
I checked in a socket update with Blocking as a property...

bfiete:
And using milliseconds. A breaking change, but I don't think too many people are doing Selects with timeouts so we might as well fix it now

damianday:
No worries thanks for that. I'll close this.

PR #630, opened 2020-10-21
Changed personal environment using user profile

obstecle:


bfiete:
It's better, but it's still an issue that it's tied to an exact NDK version.

PR #631, opened 2020-10-22
Improving README to the intuitive look and necessary info

obstecle:
What do you think?,.. it's more more nice if we are used a travis in it too.


bfiete:
Good idea with the badges, those have been incorporated.

obstecle:
i think so, at least can ++ web traffic without linking in the front.

PR #644, opened 2020-10-28
Fix formatting issue for UInt32, UInt16, UInt8 and UInt

thibmo:
Since UIt64 seems to be all fine, only these fixes are needed.


bfiete:
Did you actually find that the NumberFormatter change fixed something?
This code has been in Mono for a very long time:
https://github.com/mono/mono/blob/master/mcs/class/corlib/System/NumberFormatter.cs#L784

I know it seems wrong, but surely someone would have noticed it just not working at some point?

bfiete:
Oh, well:
```
const int Int32DefPrecision = 10;
const int UInt32DefPrecision = 10;
```
So yeah, harmless I guess.

thibmo:
> So yeah, harmless I guess.
Very harmless indeed, doesn't fix anything other then making perfectionists happy. 😉

PR #651, opened 2020-11-08
Added generic EnumToString method

M0n7y5:


bfiete:
Why not just use the `ToString` method?

M0n7y5:
@bfiete What do you mean?

bfiete:
`val.ToString(str)` works for every type. In the case of an enum, it outputs the name of the enum value.

bfiete:
Yes

M0n7y5:
It works ... nice 👍

M0n7y5:
Is this behavior in described docs?

bfiete:
No- mostly the docs only cover language features

M0n7y5:
I think it needs to be mentioned in docs or at least mention some similarities with C# so people can get the idea how Beef Enum works by looking at C# Enum docs. At least add some oneliner that will mention it in docs.

M0n7y5:
Should i close this issue?

PR #677, opened 2020-11-22
build.sh check for llvm 8 instead of 11

apg360:
It is fine in build.bat but not fine in build.sh.  
The script llvm_build.sh work with llvm 11 and not 8


bfiete:
Thanks!

Issue #679, opened 2020-11-23
25MB LLVM source code zip file instead of full LLVM git clone


apg360:
Hi,

it is very painful to git clone llvm and build it. Not mentioning that it weight 2GB download and even more generated on disk when compiling.

My question, will it work if I rather use the 25MB  LLVM source code zip file found here https://releases.llvm.org/download.html#11.0.0 ???

Second question please, would it be possible to rather use pre-built LLVM binaries instead of only source code as only option. For example on MacOS one could be able to use "clang+llvm-9.0.1-x86_64-apple-darwin.tar.xz"

Please let me know

Kinds regards

apg360:
This : https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/llvm-11.0.0.src.tar.xz

And NOT this : git clone https://github.com/llvm/llvm-project.git

Please advise

Green-Sky:
you can use a shallow clone with `--depth 1 --branch llvmorg-11.0.0` as a quick hack. downloads less than 1gig this way...

apg360:
Awesome thank you. 
As of right now, I successfully have the full git clone (2.2GB) but failing to compile it in MacOS 10.15.7 ....
cmake ../
make

Here error output  :
```
Scanning dependencies of target llvm-mt
[ 94%] Building CXX object tools/llvm-mt/CMakeFiles/llvm-mt.dir/llvm-mt.cpp.o
[ 94%] Linking CXX executable ../../bin/llvm-mt
Undefined symbols for architecture x86_64:
  "_xmlNewNs", referenced from:
      searchOrDefine(unsigned char const*, _xmlNode*) in libLLVMWindowsManifest.a(WindowsManifestMerger.cpp.o)
  "_xmlFreeNs", referenced from:
      checkAndStripPrefixes(_xmlNode*, std::__1::vector<_xmlNs*, std::__1::allocator<_xmlNs*> >&) in libLLVMWindowsManifest.a(WindowsManifestMerger.cpp.o)
  "_xmlDocSetRootElement", referenced from:
      llvm::windows_manifest::WindowsManifestMerger::WindowsManifestMergerImpl::getMergedManifest() in libLLVMWindowsManifest.a(WindowsManifestMerger.cpp.o)
  "_xmlNewDoc", referenced from:
[...]

PR #681, opened 2020-11-24
Update X86.cpp

apg360:
Is not required to do ../../ 
because cmake already does the base directory where to search for, in CMakeLists.txt :  include_directories( ../extern/llvm-project_11_0_0/llvm/include


apg360:
Juste made a fresh try and it worked fine without this change. 
I believe I made a mistake requesting this changes, it is not required.
First I will do many more build on different linux environment before asking for it again.

**Please ignore this change and do not merge it.**

PR #683, opened 2020-11-24
No need of the 2.2GB full LLVM git clone

apg360:
I make sure to keep everything at same location "llvm-project_11_0_0/llvm" so existing files are not affected.

1. extract llvm-11.0.0.src.tar.xz
2. create directory llvm-project_11_0_0
3. rename llvm-11.0.0.src to llvm and move llvm to llvm-project_11_0_0/llvm
Now everything work fine as same as git clone but it is much more light weight (25MB zip file)

Everything else remain the same. 

Please note I am not doing wget or curl to download file llvm-11.0.0.src.tar.xz before starting the process.
Because I want you to copy it in Beef repo in extern directory.
When user download Beef code, llvm-11.0.0.src.tar.xz 25MB file should already be there in extern directory.

Can download here : Click on "LLVM source code" at https://releases.llvm.org/download.html#11.0.0
Direct link : https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/llvm-11.0.0.src.tar.xz

If accepted these changes would need to be done in windows bat file too "llvm_build.bat"

### DIFF
--- a/extern/llvm_build.sh
+++ b/extern/llvm_build.sh
@@ -1,15 +1,13 @@
 #!/bin/bash
 set -e
 
+# LLVM source code, zip file 25MB : https://releases.llvm.org/download.html#11.0.0
+#https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/llvm-11.0.0.src.tar.xz 
+
 if [ ! -d llvm-project_11_0_0 ]; then
-	git clone https://github.com/llvm/llvm-project.git llvm-project_11_0_0
-
-	if [ -d llvm-project_11_0_0 ]; then
-		cd llvm-project_11_0_0
- 		git checkout llvmorg-11.0.0
-		cd ..
-	fi 
-
+	tar -xf llvm-11.0.0.src.tar.xz
+	mkdir llvm-project_11_0_0
+	mv llvm-11.0.0.src llvm-project_11_0_0/llvm
 fi
 
 if [ ! -d llvm_linux_11_0_0 ]; then


apg360:
Just tested this on Manjaro 20 and Beef compiled successfully 100%

[manj20@manjaro dist]$ ./BeefBuild -version
BeefBuild 0.42.8
[manj20@manjaro dist]$ cat /etc/lsb-release 
DISTRIB_ID=ManjaroLinux
DISTRIB_RELEASE=20.1.2
DISTRIB_CODENAME=Mikah
DISTRIB_DESCRIPTION="Manjaro Linux"
[manj20@manj20-pc dist]$

apg360:
Tested on macOS Catalina and Beef compiled successfully 100%

user@macOS dist % ./BeefBuild -version
BeefBuild 0.42.8
user@macOS dist % sw_vers -productVersion
10.15.7
user@macOS dist %

bfiete:
Including the LLVM binaries in the Beef repository will eventually cause our repository to bloat with every LLVM version.

I think the default of pulling the repository from Git is not really problematic - time time required to build, the disk space for the actual build, and the virtual memory required are by far the most common complaints. 2.2GB is a tiny fraction of the download size of a modern video game, after all...

I think the default should still be for pulling LLVM from GitHub, but I would be in support of an option to use the source archive _if_ the user has already explicitly downloaded it themselves into the extern directory. And we could update the build instructions to reflect that.

apg360:
You are the boss, your call :)

Much appreciated you took time to give feedback.
Shall I proceed to modify my change to rather : leave everything same as before but add extra check if source archive exist in extern directory, if yes then use it instead.
Is that ok for you ?

Or you have an idea/approach in mind and would rather prefer to do it yourself, so I can close this request ?

Please let me know

Kinds regards,

bfiete:
Yeah, you can modify your change as you said and resubmit. Thanks!

apg360:
Understood, thank you.

PR #685, opened 2020-11-26
use llvm-11.0.0.src.tar.xz if it exists

apg360:
User can download llvm-11.0.0.src.tar.xz into extern directory so the build will not git clone LLVM repo.


apg360:

apg360:
@bfiete  is it ok like this ?
Awaiting for your feedback please

Best regards

bfiete:
Looks good

PR #687, opened 2020-11-28
[Enhancement] lighter llvm git clone (shallow)

Green-Sky:
make llvm git clone shallow
before Xgb (idk, took too long) (~2gb compressed git download)
now 1.2gb (~130mb compressed git download)

### DIFF
--- a/extern/llvm_build.sh
+++ b/extern/llvm_build.sh
@@ -6,14 +6,8 @@ if [ ! -d llvm-project_11_0_0 ]; then
 	tar -xf llvm-11.0.0.src.tar.xz
 	mkdir llvm-project_11_0_0
 	mv llvm-11.0.0.src llvm-project_11_0_0/llvm
-  else # git clone llvm repo if llvm-11.0.0.src.tar.xz does not exists 
-	git clone https://github.com/llvm/llvm-project.git llvm-project_11_0_0
-
-	if [ -d llvm-project_11_0_0 ]; then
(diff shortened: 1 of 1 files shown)


apg360:
awesome, @Green-Sky 

Hi Sir @bfiete , what is your feedback/suggestions on this please ?

Kinds regards,

Green-Sky:
@apg360 just to make this clear: this is only for the source code, not the +20gig binaries it generates... i'm working on that one too though :wink:

apg360:
> @apg360 just to make this clear: this is only for the source code, not the +20gig binaries it generates... i'm working on that one too though 😉 (edit: like it is building the amdgpu target, which we don't need)

Indeed, it was clear for me that this resolve only the git download. From 2.2GB+ to half of it.
However I am amazed and curious when you say ~130mb compressed git download.
So on disk space after download 1.1GB, but the total download size (internet network) it will be  ~130mb because it uses compressed git download.

I am very curious and excited to test the 130mb download :D :D 

In all the cases, your changes makes the bash code in llvm_build.sh a lot more clear than before. I like that.



Regarding : <<20GB+ binaries LLVM generates after build.. i'm working on that one too though 😉 (edit: like it is building the amdgpu target, which we don't need>>
**If you do this, you will be life savior Sir !** I can't wait to see this, and will be 100% happy to be test user by redoing build from scratch to validate.

PR #688, opened 2020-11-28
[Enhancement] llvm: only build "needed" parts

Green-Sky:
Extracted currently needed parts to `llvm_targets.txt`

dir sizes are now:
```
9,6G	llvm_linux_11_0_0
1,3G	llvm_linux_rel_11_0_0
```
The debug LLVM build is now ~half the size it was before.
But this is by no means the minimal version. Hand picking dependencies (at llvm-config) might reduce the target size further.
Also, a downside of the target list is, that `make` "rechecks" every dependency for each target. (But that's still faster comparing to before). 

this commit also contains whitespace changes, sorry...

edit: @apg360 or any mac user: please check if this works for you too.


Green-Sky:
btw, i don't think most people here know but you can do:
`CMAKE_GENERATOR=Ninja bin/build.sh`
and it will just use `ninja`, which saturates the cpu more...

apg360:
> edit: @apg360 or any mac user: please check if this works for you too.

downloaded from scratch, used 25MB file llvm-11.0.0.src.tar.xz, Beef compiled successfully and helloworld project work 100%

For reference here the new size :
7,7G	         llvm_linux_11_0_0
294M	llvm_linux_rel_11_0_0

Man you did magic, thanks very much for reducing that much. this is awesome !

PR #777, opened 2020-12-31
Add IsArray and IsSizedArray to Type

xposure:


bfiete:
Ah- you only made the change in minlib, not corlib!

PR #785, opened 2021-01-04
DeleteDictionaryAndItems + String to span<char8>

xposure:


bfiete:
This is why I didn't have this before: https://github.com/beefytech/Beef/pull/277
Because I didn't think the naming of "items" clearly indicated that the keys would _not_ be deleted.
I'll add it but rename to "Values" and deprecate the "KeysAndItems" in favor of "KeysAndValues".

PR #786, opened 2021-01-04
Update llvm_build.bat

apg360:
Hi,

.bat to be like .sh
Not implemented the "llvm-11.0.0.src.tar.xz" because there is no default tar file support in windows, should we use unrar or 7zip ? .... that is a question of another day.

Kind regards

### DIFF
--- a/extern/llvm_build.bat
+++ b/extern/llvm_build.bat
@@ -1,7 +1,7 @@
 PUSHD %~dp0
 
 @IF EXIST llvm-project_11_0_0 GOTO LLVM_HAS
-git clone --depth 1 --branch llvmorg-11.0.0 https://github.com/llvm/llvm-project.git llvm-project_11_0_0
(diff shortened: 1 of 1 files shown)


apg360:
@bfiete, good morning, please have a look at this pull request.
Awaiting your feedback please.

bfiete:
Why did you remove `--config core.autocrlf=false`?

It is recommended here for Windows: https://llvm.org/docs/GettingStarted.html#getting-the-source-code-and-building-llvm

bfiete:
As for 7-zip - we could just require the user to extract the source archive themselves beforehand.

apg360:
> Why did you remove `--config core.autocrlf=false`?
> 
> It is recommended here for Windows: https://llvm.org/docs/GettingStarted.html#getting-the-source-code-and-building-llvm

@bfiete 

Very sorry, my mistake.

I just added it back now .... would you please have a look again.

Kind regards

bfiete:
Thanks

PR #873, opened 2021-01-30
Fix spelling of 'separate' and 'Separator'

bttelle:
Fix spelling of `separate`


bfiete:
Wow, I've been misspelled this one forever.

Issue #729, opened 2020-12-15
[Enhancement] Buffered FileStream


Igoorx:
Currently FileStream isn't buffered (even though there's a `bufferSize` argument in the constructor) and using it to read large files with small read calls is really inefficient and takes a long time (my test file took more than one minute to load), here's the profiling result:
![image](https://user-images.githubusercontent.com/14041768/102154120-8175f400-3e57-11eb-8199-d021d9a62718.png)

bfiete:
You're not wrong, but even with buffering I would expect the actual ReadFile calls to dominate the profile.

Igoorx:
I think I get what you mean, I didn't mean that it's inefficient because it dominates the profile, but because currently the loading of the file takes more than one minute. One way that I found to speed up things was to do a `CopyTo` to a MemoryStream and then reading the data from it, after this it took just 4 seconds to read everything, but I think this is really memory inefficient even though this works for my current case.

bfiete:
Fixed at https://github.com/beefytech/Beef/commit/4bf12e121c1e677353353c680967eea7bf43a66f

PR #875, opened 2021-01-30
Remove unnecessary parameter in StreamWriter constructor

RogueMacro:


Igoorx:
I think this parameter is unused because the buffered functionality hasn't been implemented yet (related: https://github.com/beefytech/Beef/issues/729)

bfiete:
That's correct.

RogueMacro:
Ah, that clears it up then

Issue #619, opened 2020-10-19
[BUG] Closed a dock window and it closed the entire pane


xposure:
I'm not able to reproduce this, but I accidentally clicked the `x` on the output window and it closed the entire pane (output, immediate, callstack and threads). I was debugging a memory leak when this happened if that makes a difference.

![image](https://user-images.githubusercontent.com/6292318/96471833-c2f27600-11fd-11eb-9e4c-1515b647cbbf.png)

Booklordofthedings:
[glitchprogram.zip](https://github.com/beefytech/Beef/files/5609585/glitchprogram.zip)

When opening the workspace and having the cursor in the anotherprogram panel, then debugging and running should open testclass.
When now trying to close testclass the entire panel will close.

bttelle:
This behavior is reproducible if you click on an inactive tab (say `immediate` in the example above).  The pull request above fixes this specific case.  @xposure , is it possible that you clicked on an inactive tab when this happened?

bfiete:
This does appear to be fixed by https://github.com/beefytech/Beef/commit/518c9773f367fd3e17ebc8845dc49db37db6663d

PR #879, opened 2021-02-01
Fix close of inactive IDE tab without closing pane

bttelle:
When you close an inactive IDE tab, the 'nextTab' is null instead of the
already active one. This then results in the entire pane closing even
with tabs remaining. Ensuring nextTab is always valid keeps the pane open.

This may fix issue #619 as the behavior described is similar.


bfiete:
Good fix, thanks!

PR #883, opened 2021-02-02
Fixes for incorrect SecondaryAllowClose logic

bttelle:
Existing logic relies on mWidgetWindow matching which doesn't work for inactive tabs.  Also the existing `dont save` logic is incomplete.  These commits fix several issues related to this.

### DIFF
--- a/IDE/src/IDEApp.bf
+++ b/IDE/src/IDEApp.bf
@@ -5560,30 +5560,32 @@ namespace IDE
                 });
         }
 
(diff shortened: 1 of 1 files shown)


bttelle:
@bfiete For the refactored code I left the leading spaces(!) intact. Please advise if you want all refactored code to use tabs only.

bfiete:
Great stuff, thanks!

bfiete:
But, yeah, our coding standard is for tabs but no need to refactor this right now.

PR #889, opened 2021-02-03
Fix HoverWatch handling of multiple-monitors

bttelle:
Prior to this change, 'workspaceWidth' was not retrieved with screen
coordinates so the width was sometimes wrong depending on which monitor
of a virtual screen the window was on. Sometimes the 'maxWidth'
calculated from this was negative which led to a very bad place.

### DIFF
--- a/IDE/src/ui/HoverWatch.bf
+++ b/IDE/src/ui/HoverWatch.bf
@@ -1221,13 +1221,19 @@ namespace IDE.ui
 			int workspaceY;
 			int workspaceWidth;
 			int workspaceHeight;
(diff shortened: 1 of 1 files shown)


bttelle:
image

In my setup, the virtual screen coordinates have (0,0) on the left-most monitor.  At somepoint all similar code should be tested on a setup like above where the left workspace has a negative workspaceX and screenX can be negative sometimes. `workspaceWidth - screenX + workspaceX` should still work fine.

bfiete:
Nice, thanks!

PR #902, opened 2021-02-08
Respect 'Permanent' on TabbedViews when dragging tabs out

bttelle:
TabbedViews now remain open when a single remaining tab is dragged
out of the view when the DockedWidget has 'mAutoClose = false'. This
also applies to views of secondary windows which now remain open and
empty when permanent. In this case, the full window visual no longer
participates in the drag when a single tab is dragged out.


bfiete:
Looks good, thanks!

PR #920, opened 2021-02-16
Some minor fixes for the SDL2 lib

RogueMacro:


bfiete:
This change to `CreateWindowAndRenderer` is an API-breaking change.
We should have two versions of `CreateWindowAndRenderer` - one with the old syntax and one with the new syntax, and then we can deprecate the old one to give users time to move to the new one.

Also we should add the windowpos helpers from `https://github.com/flibitijibibo/SDL2-CS/blob/master/src/SDL2.cs#L1456`

RogueMacro:
I think I have changed it as you imagined. If something's off you can edit it, like the obsolete message

PR #960, opened 2021-03-13
Fix Add From Installed project import bug

RogueMacro:
Removed VerSpec Dispose() from `DoImport()` in `InstalledProjectDialog`. This was not a necessary call as the project disposes the VerSpec automatically upon destruction. Disposing it here causes an access violation when trying to save and serialize the workspace later.


bfiete:
The issue was that `IDEApp.AddNewProjectToWorkspace` should have created a copy of verSpec instead of taking ownership of it. There are other failure cases where the `verSpec` wouldn't get disposed properly with this PR. Thanks for narrowing down the issue though!

PR #962, opened 2021-03-16
Add functions to System.IO.Stream

Moneyl:
These are some functions I needed when dealing with binary file formats. I think they would be useful additions to the class. Lemme know if you'd rather have these split into separate PRs.

- `Peek<T>()`: Reads a value from the stream without changing the stream position. If it fails to read T the stream position will still be unchanged.
- `Skip(int64 count)`: Moves the stream position forward `count` bytes. This is just `Seek(Position + count, .Absolute);` but a bit more convenient and clear what is going on for anyone reading the code.
- `WriteNullBytes(int64 count)`: Writes `count` null bytes to the stream.
- `ReadSizedString(int64 size, String output)`: Reads a string with `size` characters into `output`.
- `ReadNullTerminatedString(String output)`: Reads a null terminated string into `output`. The null terminator is read from the stream but not appended onto `output`.

The while loop at line 92 of Stream.bf is something I copied from Stream.Align to have similar style as existing corlib code. From what I understand it's supposed to be faster than writing a single null byte repeatedly. Correct me if I'm wrong.

### DIFF
--- a/BeefLibs/corlib/src/IO/Stream.bf
+++ b/BeefLibs/corlib/src/IO/Stream.bf
@@ -63,6 +63,58 @@ namespace System.IO
 		public abstract Result<int> TryWrite(Span<uint8> data);
 		public abstract void Close();
 
+		//Read value from stream without changing position. Position won't change even if it returns .Err
+		public Result<T> Peek<T>() where T : struct
+		{
+			T val = ?;
+			int size = Try!(TryRead(.((uint8*)&val, sizeof(T))));
+			Seek(Position - size, .Absolute);
+			if (size != sizeof(T))
+				return .Err;
+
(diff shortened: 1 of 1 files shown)


bfiete:
Those seem good- how do you feel about `Write(uint8 byte, int64 count)` instead of `WriteNullBytes`?
Also There's already a `ReadStrSized32` so I'd keep that naming for reading too.
Maybe `NullTerminated` is a bit long, maybe just `ReadStrC`?

Moneyl:
Those changes sound good. Pushed a commit with them.

bfiete:
Good stuff. You could always take a shot at this https://github.com/beefytech/Beef/issues/729 if you're feeling ambitious!

PR #967, opened 2021-03-30
Buffered filestream

EinBurgbauer:
I took a stab at #729 these past few days.

Everything \*should\* seem to behave just like it did before according to my tests.

A buffer size of 0 is valid and i've used it for OpenStd(), since that seemed like something that needed immediate writing. I have never used OpenStd and I'm not sure if a buffered option is useful (was thinking of something like an "buffersize = 0" arg maybe, but I don't know if std out/in interactions are actually that slow).

Let me know what you think

### DIFF
--- a/BeefLibs/corlib/src/IO/FileStream.bf
+++ b/BeefLibs/corlib/src/IO/FileStream.bf
@@ -52,7 +52,7 @@ namespace System.IO
 			return numBytesRead;
 		}
 
-		public Result<int> TryRead(Span<uint8> data, int timeoutMS)
+		public virtual Result<int> TryRead(Span<uint8> data, int timeoutMS)
 		{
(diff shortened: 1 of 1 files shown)


bfiete:
I committed some buffering that's pretty similar but pulls out the buffering part to a base class so it can be reused for other types of bufferable streams and then the delayed-write mechanics requires some changes to be able to propagate up errors from flushing methods (which even means Close can fail now).

EinBurgbauer:
That makes sense. I knew of some of the additional changes that would have to happen like the close call but wasnt sure about the details so didnt do it then. I think the it's pretty good now, thanks. I hope my stuff was of some use.

bfiete:
Yes- I'm looking forward to more PRs! :-)

PR #996, opened 2021-04-13
more buffered stream stuff

EinScott:
EDIT: this comment is outdated and for the most part not worth the read anymore.

This is a really small actually committed change, but a larger "Somethings doesn't add up here but i'm not sure which of the two are wrong" part. The commit simply resets mPos and so on in Close calls. (i mean, you can technically reuse filestreams i guess)

Now onto the other stuff. I've been running the small tests I wrote for my buffered filestream on the current filestream (i'll probably paste it below). I started looking into fixes, but ultimately just hit questions i'd rather want to ask you.

For one, seeking past the length of the stream is valid on unbuffered filestreams. Buffered streams don't allow this (which probably makes sense for most streams). On the other hand, filestreams dont let you go below 0. Buffered streams theoretically currently allow this, though i'm not sure what the use case would be (so i don't know if it should be generally forbidden or just a filestream thing).

If this is all just a filestream thing, the fix would be just to put this in BufferedFileStream to make it work like the unbuffered one.
```
public override int64 Position
{
	set
	{
		mPos = Math.Max(value, 0);
	}
}

public override Result<void> Seek(int64 pos, SeekKind seekKind = .Absolute)
{
	int64 length = Length;

	int64 newPos;
	switch (seekKind)
	{
	case .Absolute:
		newPos = pos;
	case .FromEnd:
		newPos = length + pos;
	case .Relative:
		newPos = mPos + pos;
	}

	mPos = Math.Max(newPos, 0);
	if (newPos < 0)
		return .Err;
	return .Ok;
}
```

But looking at this kind of made me notice another thing. How is Seek .FromEnd *supposed* to work? On FileStreams, it seems to do `len + seekOffset` internally or on the cpp side, so you'd have to do `Seek(-10, .FromEnd)` to stay inside the stream. But BufferedStream currently does `len - seekOffset`, so you have to do `Seek(10, .FromEnd)` to stay inside the stream (which also would make sense) - except currently it doesnt work at all because it actually does `newPos = length - pos;` without touching newPos ever again. Most other streams just fatalError on any seek that's not .Absolute, so i'm not sure about it.

I kind of made that, but I can't say i'm sure about it.
```
public override Result<void> Seek(int64 pos, SeekKind seekKind = .Absolute)
{
	int64 newPos;
	switch (seekKind)
	{
	case .Absolute:
		newPos = pos;
	case .FromEnd:
		newPos = length - pos; // May or may not need to be a +
	case .Relative:
		newPos = mPos + pos;
	}

	mPos = Math.Min(newPos, length);
	if (newPos > length)
		return .Err;
	return .Ok;
}
```

Hope that helps. Either you can fix the things or i can try and fix the things when i know more stuff :D

Lastly, here's my test. I usually abuse a diffing tool to compare the files..
```
public static void Main()
{
	FileStream fs = scope .();
	fs.Create("fs.dump", .ReadWrite);
	Test(fs);

	UnbufferedFileStream ufs = scope .();
	ufs.Create("fs2.dump", .ReadWrite);
	Test(ufs);
}

static void Test(Stream fs)
{
	fs.Write(4, 16);
	fs.Seek(10);
	fs.Write(10);
	fs.Seek(20, .Relative);

	fs.Write(10, 8000);

	fs.Seek(fs.Position-20);
	Runtime.Assert(fs.Read<uint8>() == .Ok(10));

	fs.Write<uint8>(12);
	fs.Seek(-1, .Relative);
	Runtime.Assert(fs.Read<uint8>() == .Ok(12));

	fs.Seek(-5000, .FromEnd);
	Runtime.Assert(fs.Read<uint8>() == .Ok(10));

	fs.Seek(0);
	fs.Write(255, 2);
	fs.Close();
}
```

i also had these two before, but didn't run them on this yet.
```
public static void Main()
{
	Directory.SetCurrentDirectory(Path.GetDirectoryPath(Environment.GetExecutableFilePath(.. scope .()), .. scope .()));

	BufferedFileStream bfs = scope .();
	UnbufferedFileStream fs = scope .();

	if (1 == 1)
	{
		fs.Create("fs.dump", .ReadWrite);
		bfs.Create("bfs.dump", .ReadWrite);

		Assert!();

		fs.Write(4, 16);
		bfs.Write(4, 16);

		Assert!();

		fs.Seek(10);
		bfs.Seek(10);

		Assert!();

		fs.Write(10);
		bfs.Write(10);

		Assert!();

		fs.Seek(20, .Relative);
		bfs.Seek(20, .Relative);

		Assert!();

		fs.Write(10, 8000);
		bfs.Write(10, 8000);

		Assert!();

		fs.Seek(fs.Position-20);
		bfs.Seek(-20, .Relative);

		Assert!();

		//Console.WriteLine(fs.Read<uint8>());
		Runtime.Assert(fs.Read<uint8>() == .Ok(10));
		Runtime.Assert(bfs.Read<uint8>() == .Ok(10));

		Assert!();

		fs.Write<uint8>(12);
		bfs.Write<uint8>(12);

		Assert!();

		fs.Seek(-1, .Relative);
		bfs.Seek(-1, .Relative);

		Assert!();

		Runtime.Assert(fs.Read<uint8>() == .Ok(12));
		Runtime.Assert(bfs.Read<uint8>() == .Ok(12));

		Assert!();

		fs.Seek(-5000, .FromEnd);
		bfs.Seek(-5000, .FromEnd);

		Assert!();

		Runtime.Assert(fs.Read<uint8>() == .Ok(10));
		Runtime.Assert(bfs.Read<uint8>() == .Ok(10));

		Assert!();
		
		fs.Seek(0);
		bfs.Seek(0);

		Assert!();

		fs.Write(255, 2);
		bfs.Write(255, 2);

		Assert!();

		fs.Close();
		bfs.Close();

		mixin Assert()
		{
			Runtime.Assert(fs.Position == bfs.Position);
			Runtime.Assert(fs.Length == bfs.Length);
		}
	}
	else
	{
		List<uint8> instructions = scope .();
		let count = gRand.Next(8000, 20000);

		instructions.Count = count;
		for (let j < count)
			instructions[[Unchecked]j] = (.)gRand.Next(1, 200);

		// rules
		// 10<	-> even->forward/odd->backward num * 100
		// else -> write that much

		{
			let s = scope Stopwatch(true);
			bfs.Create("bfs.dump", .Write);
			Execute(instructions, bfs);
			bfs.Close();
			Console.WriteLine(scope $"Buffered File Stream took {s.Elapsed.TotalMilliseconds}");
		}

		{
			let s = scope Stopwatch(true);
			fs.Create("fs.dump", .Write);
			Execute(instructions, fs);
			fs.Close();
			Console.WriteLine(scope $"File Stream took {s.Elapsed.TotalMilliseconds}");
		}
	}
}

static void Execute(List<uint8> instructions, Stream stream)
{
	for (let j < instructions.Count)
	{
		let inst = instructions[[Unchecked]j];
		
		if (inst < 10)
		{
			if (inst % 2 == 0)
				stream.Seek(inst * 100, .Relative);
			else TrySilent!(stream.Seek(-inst * 100, .Relative));
		}
		else
		{
			stream.Write(inst, inst);
		}
	}
}
```


EinScott:
This pull *now* does the following:

BufferedFileStream is re-openable/usable (like the unbuffered one).
BufferedStream Seek works with FromEnd.
BufferedStream Seek and Position set is unbound (most other streams allow this, and i saw no reason for this to be limited. Usually the read/write methods validate the position state and allocate more... or just error, and so it seems reasonable to not limit of length of stream by default. I don't know, maybe you have reasons i didn't think of)
BufferedFileStream Seek and Position behaves equivalent to UnbufferedFileStream's.

(So now the tests in the comment above pass)

Let me know if anything needs changing.

PR #997, opened 2021-04-14
Make Debugger only attempt to allocate hot heap when hot swapping is enabled

hunterbridges:
We are using a DLL based build and have hot swapping disabled, but every time we run in the debugger we get "Failed to reserve memory for hot swapping." This PR simply plumbs the flag through to `WinDebugger` and adds it to the `wantsHotHeap` condition in `DebugTarget::SetupTargetBinary()`.


bfiete:
Sorry it took so long! Thanks for the PR.

hunterbridges:
no worries! super glad this made it in

PR #1001, opened 2021-05-02
fixed crash when writing static unary operator

EinBurgbauer:
When writing a static unary operator, say

``public static Something operator++``

it would immediatly look for the first arg (it checked for "has it exactly one arg", but just after that) and crash on the array lookup.


bfiete:
Looks good, thanks!

PR #1013, opened 2021-05-17
Corlib additions

EinBurgbauer:
I've added some List<String> specialized functions. They are overloads of existing methods but take in StringComparison.
Also, mixins for collections with IDisposable (structs, probably) and static modifiers for classes that have no obvious reason to not be static. I was slightly afraid it might break something, but it doesnt seem to. Dictionary.ContainsValue (i dont think anyone has ever used that function) at least works for non null values now. Key also has no settable comparison yet, so i just used == for now, just like is done for keys when when inserting entries.

I'll probably do more corelib maintaining for a bit when something jumps at me. Feedback is appreciated in case i *did* break something here or you don't want some of the changes.

### DIFF
--- a/BeefLibs/corlib/src/Collections/Dictionary.bf
+++ b/BeefLibs/corlib/src/Collections/Dictionary.bf
@@ -267,7 +267,11 @@ namespace System.Collections
 			}
 			else
 			{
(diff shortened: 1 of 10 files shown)


bfiete:
Looks good!

PR #1015, opened 2021-05-18
attempt to fix explicit thiscall crash

### LINKED ISSUE #1000: [BUG] [Hard Crash] Extension method using struct with CRepr Attribute crashes compiler
This code is crashy :(
```csharp
using System;

namespace ExtensionAndThiscallBug
{
	static
	{
		public static char8* GetName(this ISomething thisPtr)
		{
			function char8*(ISomething this) vfn = (.)thisPtr.VT[0];

			return vfn(thisPtr);
		}
	}

	[CRepr] // < -- Apparently this is the issue
	struct ISomething
	{
		public void** VT;
	}

	class Program
	{
		{
			ISomething something = default;

			let name = something.GetName();

			return 0;
		}
	}
}
```

EinBurgbauer:
an attempt to fix #1000

As far as I could tell, the issue arised because GenIRFunctionInfo - used to get the number of args as well - will on index -1 get the info of the 0th arg if it's an explicit "this" and treating it like that, but then processing it again on index 0. Consequently, SetupIRMethod thinks the function has two args, and does an out of bounds array access.

My usual projects compile fine and Beef also bootstraps normally (nothing caught fire), though, you know, i really dont have omnisense in this codebase and am in fact not exactly a compiler pro.

### DIFF
--- a/IDEHelper/Compiler/BfResolvedTypeUtils.cpp
+++ b/IDEHelper/Compiler/BfResolvedTypeUtils.cpp
@@ -1208,7 +1208,8 @@ void BfMethodInstance::GetIRFunctionInfo(BfModule* module, BfIRType& returnType,
 	{
 		returnType = module->mBfIRBuilder->MapType(mReturnType);
 	}	
-	
+
+	bool hasExplicitThis = false;
 	for (int paramIdx = -1; paramIdx < GetParamCount(); paramIdx++)
 	{
 		BfType* checkType = NULL;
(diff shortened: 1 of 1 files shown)


bfiete:
This broke the test suite - I disabled the change for now.

The test project is in IDEHelper\Tests. You can also run bin\test_build.bat after bin\build.bat.
Ideally we'd add a test that triggers https://github.com/beefytech/Beef/issues/1000 to that test project as well.

EinBurgbauer:
I see, I'll think of that in future, thanks.

PR #1016, opened 2021-05-20
fix thiscall crash for real

### LINKED ISSUE #1000: [BUG] [Hard Crash] Extension method using struct with CRepr Attribute crashes compiler

EinBurgbauer:
As it turns out, there thankfully wasn't another issue, but i was just being incompentent. There was already a check in place at the end of the loop to skip index 0 on explicit this, but this case just continue;'d past it, so i now moved that check to where the old fix was. I also included a test that crashed before, but doesnt now.

Building LibC for tests required me to dig up and install windows SDK 10.0.18362.0 very insistantly, but i wasn't really prepared for what came after, i mean- beefperf??? ide scripts?? i had no idea that existed (need to look at it at some point i guess) :D are ide scripts just for internal testing, what are their capabilities?

Anyways, should fix #1000 for real this time.

### DIFF
--- a/IDEHelper/Compiler/BfResolvedTypeUtils.cpp
+++ b/IDEHelper/Compiler/BfResolvedTypeUtils.cpp
@@ -1227,8 +1227,7 @@ void BfMethodInstance::GetIRFunctionInfo(BfModule* module, BfIRType& returnType,
 				{
 					checkType = GetParamType(0);
 
-					//TODO(BCF): Breaks tests
-					//hasExplicitThis = true;
+					hasExplicitThis = true;
 				}
 				else
 					checkType = GetOwner();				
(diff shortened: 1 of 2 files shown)


bfiete:
BeefPerf is pretty cool - it's similar in design to http://www.radgametools.com/telemetry.htm

The scripting support is very basic - it's actually the same system that the post-build project command system uses in the IDE, which is mostly just a list of reflected methods to call. It's for building tests for incremental compilation scenarios, the debugger, hot compilation, etc. It doesn't even support looping and it's not even turing-complete, though.

PR #1017, opened 2021-05-21
fix import project from installed

EinBurgbauer:
searching a project in the import for installed dialogue would display filtered results, but still select from the unfiltered list on click, because refiltering didn't clear the filtered list before adding.


bfiete:
Thanks!

PR #1019, opened 2021-05-26
Builtin attribs fix

EinBurgbauer:
stuff like [System.InlineAttribute] wont actually work, but is sometimes more or less needed if you're emitting code.

I was also today years old when i learned that builtin attributes actually just do nothing when typealiased. I mean, i guess there is mostly no use in it, but it also means that this didn't work for half a year and i somehow never noticed.
```
[AttributeUsage(.Method)]
struct DebugCallAttribute : Attribute {} // Useless, basically

#if DEBUG
typealias DebugOnlyAttribute = DebugCallAttribute;
#else
typealias DebugOnlyAttribute = SkipCallAttribute;
#endif
```
still, arguably, this is one of the very few sensical applications of typealiasing builtin attributes.

### DIFF
--- a/IDEHelper/Compiler/BfAst.cpp
+++ b/IDEHelper/Compiler/BfAst.cpp
@@ -1195,6 +1195,23 @@ bool BfTypeReference::IsTypeDefTypeReference()
 	return IsA<BfNamedTypeReference>() || IsA<BfDirectStrTypeReference>() || IsA<BfDirectTypeDefReference>();
 }
 
+String BfTypeReference::ToCleanAttributeString()
+{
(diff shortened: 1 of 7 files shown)


bfiete:
Yeah- there's a few ways attributes are handled. Some attributes are processed during parsing (and attached to a BfTypeDef/BfMethodDef), and then there's things like `BfModule::ProcessTypeInstCustomAttributes` which is handled during type instantiation and those should actually work with type aliases...  And then there's some cases where we use `BfCustomAttributes::Contains` which should also work with type alises...

EinBurgbauer:
From what i tested at least SkipCall doesnt seem to work, i'll have a look at those, thanks

PR #1026, opened 2021-05-30
[WiP] [corlib] Add System.Caching (BeefLang port of System.Runtime.Caching)

thibmo:
I noticed that corlib is still missing the Caching utils that are available in .net, which are rather useful.
They allow us to cache items, but also helps us prevent double-triggering events where it's really not needed.

Could also be used at a later stade for things like SQL query caching, web response caching, etc.

I did have to add a simplistic Timer (System.Threading.ConsoleTimer) to get things to work properly.

Improvements are always welcome, I'll be porting it over-time and only enough to have a simple implementation functional, which can then be built upon.


bfiete:
What's the name "ConsoleTimer" supposed to mean? It looks more like a "PeriodicCallback". Is ConsoleTimer some type of terminology used somewhere in this `System.Runtime.Caching`?

A few things I noticed:
1) This should probably be in a separate library rather than put directly into corlib
2) It seems you could have just done `_cancelEvent(_interval);` in the thread loop, but you're using a `StopWatch` presumably because you don't trust the wait interval of the event to be accurate... I assume you meant to adjust the next wait time to account f or any overage or underage from the previous _measured_ wait time, but it doesn't appear the math actually accomplishes that. It looks like you wanted something like `waitInterval = _interval + (_interval - lastProcTime);`. Also there's no real reason to have a lock around the _interval read -  if you're worried about a non-atomic 64-bit update on a 32-bit system... well, just use a 32-bit _interval anyway and you never have to worry about that.
3) It seems `_threadExited` is redundant since you can always just wait on `_thread.Join` for the same effect.
4) The `if (interval < -1) Runtime.FatalError(...)` is usually handled by a `Runtime.Assert(interval >= 0)` - also remember that developers always see the source code in a crash, and the code itself documents the requirement better than hand-written text.
5) Why bother with the two constructors? If 64-bit intervals are not required the just have the `int` one.

thibmo:
Thanks for the comments, renaming it is rather easy, in the source ref they use Timer to check cache expiry on intervals.
Since we don't have a Timer yet I made this as a way to still have something alike a Timer.

The StopWatch is used to offset the next interval with the time it took to execute the callback, so we don't have situations where constantly miss the next interval by an ever-increasing amount of time.

I'll put more comments and simplify things more today, after that I'll also start with the actual reason for making this PR, System.Caching. :)

thibmo:
Pushing it as-is right now, it's not nearly close to being done, but better to have it in the repo.

bfiete:
This looks good. Also, you do know you can freely use code from Microsoft's implementation at https://github.com/microsoft/referencesource/tree/master/System.Runtime.Caching/System/Caching

I used some code from that repository even for the corlib.

But anyway- I know this is a first-party library for Microsoft but I don't have the resources to maintain libraries here outside the corlib itself. This code would be best exposed as an external third-party library. When we have the package manager working properly then external libraries will be just as easy to use as "official" first-party libraries, and then I'll even remove some of the current first-party libraries (ie: curl, SDL2) out of the main repository and out of the standard Beef distribution.

thibmo:
> This looks good. Also, you do know you can freely use code from Microsoft's implementation at https://github.com/microsoft/referencesource/tree/master/System.Runtime.Caching/System/Caching
> 

Ah yes, also used that, but the website version where it's easier to search for specific methods/classes https://referencesource.microsoft.com/

Sadly it's missing certain critical classes which I had to either search for or make up.

I'll pull it over to a separate lib, that should be doable. If some assistance is needed with features, feel free to drop hints. :)

thibmo:
Moved to https://github.com/thibmo/System.Caching

PR #1077, opened 2021-06-21
Fix a thread safety issue in TimeZoneInfo

### LINKED ISSUE #1056: Thread-safety problem with TimeZoneInfo
The function `TimeZoneInfo.CachedData.GetOneYearLoadFromUtc` isn't thread-safe and this can cause `DateTime.Now` to leak when running in multi-thread.

To solve this issue, I suggest to replace the line 256 of `TimeZoneInfo.bf` with the following code.
```
					if (Interlocked.CompareExchange(ref m_oneYearLocalFromUtc, null, oneYearLocFromUtc) != null)
					{
						delete oneYearLocFromUtc;
						oneYearLocFromUtc = m_oneYearLocalFromUtc;
					}
```

disarray2077:
Fixes #1056

This is the fix that I suggested in the issue #1056.

Basically, before exchanging the value it checks if the value already was changed by another thread and if it was, we just pick it and delete what we constructed in the current thread.

### DIFF
--- a/BeefLibs/corlib/src/TimeZoneInfo.bf
+++ b/BeefLibs/corlib/src/TimeZoneInfo.bf
@@ -253,7 +253,10 @@ namespace System {
 					if (rule != null)
 						rule = rule.Clone();
                     oneYearLocFromUtc = new OffsetAndRule(year, currentYear.BaseUtcOffset, rule);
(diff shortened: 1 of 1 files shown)


bfiete:
Looks good.

PR #1083, opened 2021-06-24
Add Remove method to Dictionary Enumerators

disarray2077:
Sorry if I missed something, but looking at the implementation of both `Dictionary.Remove` and the Dictionary Enumerators, I didn't see any reason to justify a `Remove` method (like the List one) not existing, so I suggest to add it with this pull-request as this is really useful.


bfiete:
I added this as-is, but I think we could make the actual remove faster since we already have the entry index - at least the hashing and comparisons can be avoided, if not the entire loop.

PR #1089, opened 2021-07-02
Partial support for scientific notation in float.Parse()

Moneyl:
Adds scientific notation to `float.Parse()`. Only supports syntax where you prefix the exponent with `e` or `E`. Doesn't support `base * 10 ^ exp` syntax. There's some examples of supported syntax in `/IDE/Tests/Test1/src/Parse.bf`.


bfiete:
I put this in at 1efdb334b42c9c25dfa39b4b2a12d41f7238be8c
Also- the `IDE/Tests` are for scripted IDE testing. The normal compiler tests are at `C:\Beef\IDEHelper\Tests`, so I added the tests there.

Thanks!

Issue #1058, opened 2021-06-19
File is erased if you try to save when the disk is full


disarray2077:
If you try to save a file when the disk is full, an assert failure occurs, this causes the save to only erase the contents of the file and if there is no recovery file, your file is gone.

disarray2077:
I found where the assert failure (`Unhandled error in result`) happens, it's in this line: https://github.com/beefytech/Beef/blob/e5f92fb21b327b8d5c89da4a413d509a5169db20/BeefLibs/corlib/src/IO/BufferedStream.bf#L154
I would do a PR to fix it, but I'm not sure what to do, this isn't an error that should be ignored but the current code seems to want to ignore all errors, and I have no idea why it was written like this.

bfiete:
Just an oversight. An `else return .Err;` would work there.

disarray2077:
This ended up being more complicated than I expected, but the pull request that I did seems to solve the problems related with this issue 👍

PR #1130, opened 2021-08-01
Fix issues with the methods for writing files

disarray2077:
Fixes https://github.com/beefytech/Beef/issues/1058

This pull request is composed of multiples changes:
- All Write methods from File class where changed to use `UnbufferedFileStream`, as I guess `FileStream` doesn't really have any benefit in this case and `FileStream` needs special handling of the `Close` method because of it's buffered nature, so it's not worth using it unless necessary, or when the `Close` wouldn't cause problems, like when we are reading only.
- Added create kind `OpenAlways` and changed `OpenOrCreate` to use it, this is necessary because `CreateAlways` Truncates the file on Windows. _(I didn't test this on Linux, but it looks right from what I could read from the source code)_
- Added SeekUnderlying to FileStream to reduce code duplication.
- Implement `FileStream.SetLength`, this is a way to make sure that we have the space to write a file, thus avoiding problems when the disk is full.
- Improved WriteTextFile safety by using the `FileStream.SetLength` function.
- Fixed two error propagation issues in BufferedStream.

### DIFF
--- a/BeefLibs/Beefy2D/src/Utils.bf
+++ b/BeefLibs/Beefy2D/src/Utils.bf
@@ -295,11 +295,13 @@ namespace Beefy
 
         public static Result<void> WriteTextFile(StringView path, StringView text)
         {
-			var stream = scope FileStream();
-			if (stream.Create(path) case .Err)
-			{
+			var stream = scope UnbufferedFileStream();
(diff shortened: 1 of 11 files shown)


bfiete:
This all looks like good stuff. Thanks.

PR #1197, opened 2021-10-31
Added button to toggle case sensitivity in quick find and quick replace

MineGame159:
By default case sensitivity is disabled and persists between actions as long as the IDE didn't restart (it is stored in a static variable).
I wasn't sure how to handle the color so I just hardcoded it. If there are any changes I could make I will be happy to learn.
image
image

### DIFF
--- a/BeefLibs/Beefy2D/src/theme/dark/DarkButton.bf
+++ b/BeefLibs/Beefy2D/src/theme/dark/DarkButton.bf
@@ -75,14 +75,16 @@ namespace Beefy.theme.dark
 
             g.SetFont(DarkTheme.sDarkTheme.mSmallFont);
             if (mLabel != null)
-            {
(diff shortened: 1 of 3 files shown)


bfiete:
I'm planning on checking out these PRs after the next stable release.

bfiete:
image

Compare to Visual Studio's Quick Find, which includes features I would find necessary:
1) Includes the Whole Word toggle. The UI layout should also include space for a future Regular Expression toggle
2) You can see whether or not the toggle is actually set. In your mockup I can't even tell you whether the blue or the white text means "on" or "off"
3) "Aa" is the industry standard, not "Ww"

In addition, this violates our own UI standard of not persisting options/toggles in tool windows. Compare to the our 'Find in Files' dialog: there is a "Prev" button if you want to go back to previously-set options, but otherwise you get a standard known set of options. This is important for being able to quickly and reliably use a tool without having to visually scan whether the wrong options have been set from a previous usage.

Issue #981, opened 2021-04-02
[Enhancement] Opt.for alternate tab ordering


marsej:
I'd like if the "most time spent open"-tabs would be on the left side of editor tab bar since their location doesn't change.

eg. if Program.bf has been open 10 seconds, then opening Another.bf, Another.bf would be on right side of Program.bf since it's been open only 0 seconds.

In 0.43.0 If new file is opened, it opens to the left-most tab. This is kinda like if you played piano and every new key pressed would alter locations of other keys.

marsej:
Thanks for addressing this.

PR #1252, opened 2021-12-01
[IDE] Added setting to open new tabs on right instead of left

blat-blatnik:
Title should be pretty self explanatory. 

There's now a new setting to change whether new tabs appear on the left, or on the right.

Tabs opening on the left bugged me a little bit since I'm used to Visual Studio, which opens tabs on the right (by default at least). Also it seemed like an easy thing to add for my first contribution :)

For now I've left the old behavior as the default.

Let me know if something's not up to snuff.

screenshot

I see now that someone else requested something like this in #981


bfiete:
There's a couple other viable options too: Sublime Text (for example) inserts new tabs to the right of the active tab...

Issue #1176, opened 2021-10-27
`Dbg_ObjectAlloc` should use the workspace `maxStackTraceDepth` by default


disarray2077:
`Type.CreateObject` creates an object using `Dbg_ObjectAlloc` with `maxStackTraceDepth` equal to 1, but it should rather follow the workspace configuration.
https://github.com/beefytech/Beef/blob/f7853fc3481edf55ae7c1b954b9c675e61f08fd1/BeefLibs/corlib/src/Type.bf#L1190
https://github.com/beefytech/Beef/blob/f5a863d55a35ed85e99ff612305aafe3e14b3825/BeefLibs/corlib/src/Reflection/TypeInstance.bf#L126
https://github.com/beefytech/Beef/blob/c10565678ebedf7bfc237e25ae9f661ed98a03c2/IDE/mintest/minlib/src/System/Type.bf#L1120

PR #1260, opened 2021-12-04
Expose `AllocStackCount` to corlib & Use it in Dbg_ObjectAlloc

disarray2077:
Closes https://github.com/beefytech/Beef/issues/1176


disarray2077:
Damn, a unintended change slipped in the pull request

bfiete:
Note that this doesn't respect Distinct Build Options (ie: the ability to set one partial type's alloc stack count to a higher level). To do that we'd have to encode that option in the type's reflection data...

disarray2077:
> Note that this doesn't respect Distinct Build Options (ie: the ability to set one partial type's alloc stack count to a higher level). To do that we'd have to encode that option in the type's reflection data...

Oh I see, so that `typeOptions->mAllocStackTraceDepth` was for this functionality... I guess this pull request will remain as a draft for now 😅
I will try to look into this.

bfiete:
We can do this without even increasing the size of the type data- there's padding available at the end of `System.Type`, so we can use one byte for this and reflection data remains the same size.

disarray2077:
Done, I had to limit the value to 255 to fit a byte tho, does this look right @bfiete?

Issue #976, opened 2021-04-02
[Bug] SDLApp.bf handles volume in a way that causes momentarily louder audio when firing laser in SpaceGame


marsej:
SpaceGame PlaySound doesn't mix sounds consistently. This can be noted when there are multiple simultaneous sounds supposed to be played.

REPRO: I noticed this issue by lowering the volumes in the SpaceGame in various files where volume was set for each sound. The change lowered the volumes but once in a while it would play the sounds at full volume. 

Even without simultaneous sounds, the laser sound plays unevenly but it's bit harder to notice.

marsej:
The problem turned out to be that in PlaySound channel -1 allocates a new channel for each instance of sound played. If there are many explosion sounds going on, they all get their own channel and get summed up to form some very loud explosions. 

I made following mods to solve this:
```
SDLApp.BF
		public void PlaySound(Sound sound, int ch = -1, float volume = 1.0f, float pan = 0.5f)
		{
			if (sound == null)
				return;

			int32 channel = SDLMixer.PlayChannel(ch, sound.mChunk, 0);
			//SDLMixer.SetPanning()
			SDLMixer.Volume(channel, (int32)(volume * 128));
		}
```

Then I had the 
EnemyGolaith's.bf call
					gGameApp.PlaySound(Sounds.sExplode, 2, 0.45f, 1.0f);
EnemyProjectile.bf call
				gApp.PlaySound(Sounds.sExplode, 2, 0.25f, 1.5f);
GameApp.bf call
				PlaySound(Sounds.sLaser, 1, 0.05f);
HeroBullet.bf call
							gGameApp.PlaySound(Sounds.sExplode, 3, 0.2f, 1.5f);


This way they ended up being sounded on same channel as before, preventing excessive summing.

PR #1261, opened 2021-12-07
PlaySound sets SDL volume for the sound played instead of the app

marsej:
I didn't include these comments in the change
This pull closes issue https://github.com/beefytech/Beef/issues/976

SDLMixer.VolumeChunk(sound.mChunk, (int32)(volume * 128)); // sets volume for this sound only
int32 channel = SDLMixer.PlayChannel(-1, sound.mChunk, 0);
//SDLMixer.SetPanning()
SDLMixer.Volume(channel, (int32)(volume * 128));
//SDLMixer.Volume(channel, (int32)(volume * 128)); // sets app volume

### DIFF
--- a/BeefLibs/SDL2/src/SDLApp.bf
+++ b/BeefLibs/SDL2/src/SDLApp.bf
@@ -191,9 +191,9 @@ namespace SDL2
 			if (sound == null)
 				return;
 
+			SDLMixer.VolumeChunk(sound.mChunk, (int32)(volume * 128));
 			int32 channel = SDLMixer.PlayChannel(-1, sound.mChunk, 0);
 			//SDLMixer.SetPanning()
-			SDLMixer.Volume(channel, (int32)(volume * 128));
 		}
 
 		public void Render()


bfiete:
The volume is supposed to be for the specific instance being played, not the chunk - this change would also change the volume of every other instance of the sound being played, as well.

`SDLMixer.Volume` does not change the app volume, it changes a specific channel volume- ie: the sound instance.

It seems the proper fix would be to manage channels allocations directly instead - so we have a specific channel to set the volume on _before_ the sound is played.

marsej:
Quick search shows someone having exactly same issue in similar situation:
https://stackoverflow.com/questions/63220371/sdl-mixer-audio-randomly-gets-louder

I did a test and found that the following also fixes the volume spiking (idea is that each level of volume has its own channel):
```
		public void PlaySound(Sound sound, float volume = 1.0f, float pan = 0.5f)
		{
			if (sound == null)
				return;
			int32 channel = SDLMixer.PlayChannel((int32)(volume * 128), sound.mChunk, 0);

			//SDLMixer.SetPanning()
			SDLMixer.Volume((int32)(volume * 128), (int32)(volume * 128));
		}

		public void Run()
		{
			int32 ac=SDLMixer.AllocateChannels(129);
...
```
The spiking volume (using channel "-1") occurs when there's a decent amount of lasers and explosions going off simultaneously, it doesn't repro very consistently but when it does, the volume is as if SDLMixer.Volume was not called here, so very very loud.

Thoughts on this approach?

bfiete:
You're not allocating a channel to play on, you're using a channel id of `(int32)(volume * 128)`, which will interrupt a previously-playing sound on whatever that channel id ends up being.

marsej:
How about this workaround?

```
			SDLMixer.VolumeChunk(sound.mChunk, (int32)(0)); // initially silence
			int32 channel = SDLMixer.PlayChannel(-1, sound.mChunk, 0); // just to get a channel
			SDLMixer.Volume(channel, (int32)(volume * 128));
			SDLMixer.VolumeChunk(sound.mChunk, (int32)(128)); // now that channel vol is set, restore sample volume
			SDLMixer.PlayChannel(channel, sound.mChunk, 0);
```

It seems to work (no more vol spikes), the only mystery is why I have to set the volume really really low (0.05f) for EnemyProjectile to match same sound (sExplode) which is set to 0.16f for the HeroBullet.

bfiete:
The real fix is just handling the `-1` channel allocation error result from `SDLMixer.PlayChannel`.
Fixed at 195c705a46e685f80aed65b9c75996e41f3b9f18

Your change will not allow for multiple explosion sounds to overlap at all - you'll chop them off and restart them, which is.. not good.

bfiete:
Well- this is _part_ of the real fix. There's still technically a race condition between playing the sound and setting the volume where part of the sound could be played at the old volume, but fixing that would requiring handling channel allocations manually which was outside the scope of this sample SDL app example.

marsej:
Good enough fix thank you. I can still repro high loudness but only with bit unrealistic usage (making the enemy shoot at almost continuous rate, which is the case for my projectile aiming/firing-solution debug scenario).

PR #1262, opened 2021-12-07
Sdl white flash fix

marsej:
When starting up SpaceGame in SDL Accelerated mode, profiling says 600 ms is spent in WinVerifyTrust. And this is when the SDL has initialized the window to default white background. To the user this appears as a white flash before the games black background appears.

I changed the code to hide the window until it's background color is set to black.


marsej:
I have no idea how could I have avoided the previous SDL sound volume fix commit appear in this second pull request. My wish/intention was to have each change to the same file be a different pull request but not sure if that's how you want it and how to implement it using the git command line.

PR #1263, opened 2021-12-07
Changed SDL initial bg color from white to black to remove white flas…

marsej:
…h effect with some GPUs

Changed SDL initial bg color from white to black to remove white flash effect with some GPUs

My NVIDIA user mode driver spends 600-700 ms in WinVerifyTrust when using .Accelerated SDL initialization.
The problem is seen as a brief white flash/strobe effect before the game loads.

If the app was something like white background notepad, then changing that color to white or removing this commit would of course be reasonable but games rarely use 100% white as background so black is a better default.

beef spacegame startup profile december
beef spacegame startup profile december2


bfiete:
If someone had a bright-colored game then they would now say that the screen flashes black. It doesn't really solve the underlying issue.

It seems a more proper fix would be to delay display of the window until after the first _actual_ screen draw.

PR #1266, opened 2021-12-11
Added editor option for // -style comments in certain cases

marsej:
If VS is installed and C# was selected as configuration template then CTRL+K+C works differently than in VS C++. 
A new option under Editor ("Toggle comment alternate style") was added that more closely mimics the VS C# style commenting behavior.

### DIFF
--- a/IDE/src/Commands.bf
+++ b/IDE/src/Commands.bf
@@ -194,6 +194,8 @@ namespace IDE
 			Add("Close Document", new () => { gApp.[Friend]TryCloseCurrentDocument(); });
 			Add("Close Panel", new () => { gApp.[Friend]TryCloseCurrentPanel(); });
 			Add("Close Workspace", new => gApp.[Friend]Cmd_CloseWorkspaceAndSetupNew);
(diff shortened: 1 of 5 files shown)


marsej:
The "commenting blank now does nothing" commit contains important change so it should be applied if the ToggleCommentAlt is applied.

bfiete:
I have a different proposal for the implementation:
We rename the "Comment Selection" to "Comment Block" and then we add a new "Comment Lines" command that performs the `//` commenting style.

Then you can switch between styles with just a key bind - or the user can bind each style to a different key sequence.

Also, the "Uncomment Selection" command doesn't yet recognize and uncomment the `//` style.

marsej:
For some reason git says "added 6 commits 6 days ago" when 5 of those commits above were just added. 
Uncommenting should now work, the new binds for the new CommentLines and CommentBlocks are in "Add new key-defaults" commit.

marsej:
Getting a crash going to Settings/Editor. Tried Reset settings but didn't help. IDK whats up (yet)..

PR #1276, opened 2021-12-14
Optimize the remove method of dictionary enumerator

disarray2077:
This is an optimization as suggested in https://github.com/beefytech/Beef/pull/1083, It only avoids hashing and comparisons, I don't think we can avoid the loop (without a memory overhead) since we can only know the index of the last entry if we loop through the entries list.


bfiete:
Looks good.

PR #1277, opened 2021-12-15
Fuzzy string matiching for autocomplete

aharabada:
Fuzzy string matching is one of the features I have been missing the most compared to other IDEs like Visual Studio.
I implemented it using [this](https://github.com/forrestthewoods/lib_fts/blob/master/code/fts_fuzzy_match.h) library which I modified a bit so that it works with UTF-8 strings.

Here is a screenshot of the fuzzy autocomplete "in action":
example

To be honest the way I implemented it on the C++ side porbably isn't the best... but I thought it was good enough to present as a proof of concept.


bfiete:
Have you done any perf testing? Testing against something like https://github.com/blat-blatnik/BeefWin32 is useful, as it creates a `Win32.Win32` class with many, many entries that need to be filtered through in a reasonable speed.

aharabada:
I've finally done some perfomance testing.
At first the performance results were pretty bad (autocomplete for "Win32." took about 3.5s), but I've pushed a commit that improved the performance quite a bit.

To test the performance I started a Stopwatch at the start of `SourceViewPanel.DoAutoComplete` and stopped it at the end of `AutoComplete.SetInfo`.

In the table below are unscientific performance numbers meassured for a couple of "queries" I chose for no particular reason.
Notice the column "Relative perf per match" which was calculated using this formula: (classic time / matches) / (fuzzy time / matches). If the number in that column is larger than 1 fuzzy matching is faster per match than the classic matching.

| Autocomplete text    | Classic matching  | Fuzzy matching    | Relative perf per match |
| -------------------- | ----------------- | ----------------- | ----------------------- |
| "W"                  | ~15ms (3 matches) | ~15ms (8 matches) | 2.666                   |
| "Win32."             | ~1.3s (133091)    | ~1.45s (133091)   | 0.897                   |
| "Win32.A"            | ~30ms (3356)      | ~850ms (100111)   | 1.053                   |
| "Win32.As"           | ~50ms (7295)      | ~300ms (51822)    | 1.184                   |
| "Win32.A2"           | ~50ms (179)       | ~80ms  (3733)     | 13.034                  |
| "Win32.D"            | ~80ms (23040)     | ~700ms (87793)    | 0.435                   |
| "Win32.D3"           | ~55ms (3869)      | ~75ms  (5326)     | 1.009                   |
| "Win32.D3D"          | ~55ms (3835)      | ~75ms  (4337)     | 0.829                   |
| "Win32.D3DTEXTURE2D" | ~55ms (0)         | ~70ms  (16)       | ---                     |
| "Win32.ID3D11Device" | ~50ms (12)        | ~85ms  (19)       | 0.931                   |

Overall I think the performance is acceptable. When it takes unconfortably long for the results to arrive we get way too many results anyways. Besides, the bottleneck doesn't seem to be the fuzzy search but rather the vast amount of results.

bfiete:
Hm, that's better than I expected. I do think should have a preference setting to toggle between classic & fuzzy autocomplete, though.

aharabada:
I've added the "Fuzzy Autocomplete" option to toggle between classic & fuzzy autocomplete. By default it uses classic autocomplete.

aharabada:
Do you think it would make sense to highlight the matched characters for classic autocomplete too? And perhaps add a toggle for that.
I didn't do it for now as I think it's not as useful for classic autocomplete as it is for fuzzy autocomplete.

bfiete:
Turns out autocomplete was using a terrible hash function. There's a new one at 71745110a51740014b113dd619612535dc13f3fa which dramatically improves speed of both matching methods

Issue #1064, opened 2021-06-19
Go to definition goes to base definition of property instead of overrided one


disarray2077:
If you click to Go to Definition of a overrided property, it goes to the base definition of the property instead of the overrided one.
Example:
```
		public class MyBaseClass
		{
			public virtual int A => 0;
		}

		public class MyClass : MyBaseClass
		{
			public override int A => 1;

			public void Test()
			{
				Console.WriteLine(A); // Clicking "Go to Definition" in "A" goes to the base definition.
			}
		}
```
_(PS: From what I saw, this only happens with properties)_

PR #1304, opened 2021-12-23
Fix Property Go To Definition

disarray2077:
Fixes https://github.com/beefytech/Beef/issues/1064

I hope this doesn't break anything, as it looks like this behavior was intentional.


bfiete:
It was intentional but I don't remember why. Do you have a case where the current behavior is broken?

disarray2077:
The code in the issue https://github.com/beefytech/Beef/issues/1064 shows what is wrong with it.
(It goes to the base property instead of the overridden property)

bfiete:
Oops- yeah, I should have seen that in the text up there.

PR #1308, opened 2021-12-27
Add IsDefined<T>, GetValues<T> and GetNames<T> to Enum class

disarray2077:
... and removed an obsolete commented-out code.

### DIFF
--- a/BeefLibs/corlib/src/Enum.bf
+++ b/BeefLibs/corlib/src/Enum.bf
@@ -30,12 +30,118 @@ namespace System
 			return .Err;
 		}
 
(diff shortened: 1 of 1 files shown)


bfiete:
There's currently an issue where this will also enumerate over non-case fields such as the discriminator in enums with payloads, or user-define constants as in the case of:
```C#
enum TestENum
{
	case A;
	case B;
	case C;
	const int mD = 123;
}
```

I added a `EnumCase` field flag that can be checked to filter out non-case fields

disarray2077:
I didn't even think about this kind of enum 😅, I guess it should work correctly now.

PR #1357, opened 2022-01-07
Add option to highlight current line in editor

blat-blatnik:
Makes it easier to keep track of where you're currently editing.

screenshot

I tried to make it look like it does in Visual Studio.

reference

I only have it enabled for source code edit panels - so it won't show up in the immediate window for example. There are also options to disable it (it's enabled by default).


bfiete:
I disabled this by default, but I do now hilite the line number on the left which I think partially achieves the same results less intrusively.

PR #1368, opened 2022-01-09
Toggle breakpoint at begin scopes

marsej:
Added new command that toggles breakpoint(s) at begin of scopes. Not bound by default.

### DIFF
--- a/IDE/src/Commands.bf
+++ b/IDE/src/Commands.bf
@@ -205,6 +205,7 @@ namespace IDE
 			Add("Breakpoint Symbol", new () => { gApp.mBreakpointPanel.AddSymbolBreakpoint(gApp.[Friend]GetCurrentWindow()); });
 			Add("Breakpoint Toggle Thread", new => gApp.[Friend]ToggleThreadBreakpoint, .Editor);
 			Add("Breakpoint Toggle", new => gApp.[Friend]ToggleBreakpoint, .Editor);
+			Add("Breakpoint Toggle at begin scopes", new => gApp.[Friend]ToggleBeginScopeBreakpoints, .Editor);
 			Add("Build Workspace", new => gApp.[Friend]Compile);
 			Add("Cancel Build", new => gApp.[Friend]CancelBuild);
 			Add("Clean Beef", new => gApp.Cmd_CleanBeef);
--- a/IDE/src/IDEApp.bf
+++ b/IDE/src/IDEApp.bf
@@ -4505,6 +4505,16 @@ namespace IDE
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


marsej:
I recommend using this only both are satisfied:
a) when no existing breakpoints exist at begin of scopes in current active file
b) you receive an unknown project with large files and want to break where methods start executing. 

A similar feature was in another debugger 20 years ago.

marsej:
I should note that for some reason bp's at { get moved down to the first command unlike in VS where they stay on {'s.
For this reason I added another loop to skip lines beginning with / or whitespace/blank lines. So it correctly now removes BP's even after running the project.

bfiete:
We do explicitly ensure there's always an instruction (even just a NOP) at the closing brace of a block - that's so if you have something like
```C#
{
  int a = GetVal();
}
```
then you can actually step over the variable declaration and actually see its value (since it's still in scope). That's something Visual C++ (not clang) do not do.

C# does that, however, and also ensures an instruction on the opening brace. We could do that as well, but I didn't really see the use.

What _is_ the use case of adding a breakpoint to the start of every single block?

marsej:
When you know a certain large file has something you want to understand better and is being called from elsewhere but you don't know where to put a BP. Using API logger would be a 'sort of' alternative to this but I don't see any API logging/tracing features, are there any?

Also I didn't see any documentation on how to use "symbol breakpoints", like, do you put a filename as a symbol breakpoint to break on entry to anywhere in that file?

bfiete:
Symbol breakpoints let you manually specify a symbol name. IE: `CreateWindowExA`.

For your use case you probably actually want to have a logging breakpoint rather than a normal breakpoint. With "Log String" you can add params in there as well, ie: `Entering API.Init({param1}, {param2})`

I don't even know if there's a way to do something like that for C#, is there?

marsej:
This idea doesn't come from C# debugger but OllyDbg which had some advantages over VS debugger back in 2002. With OllyDbg you can get a list of all symbols and rapidly add BPs or log/trace on entry to all of them by first clicking "search for all intermodular calls" then selecting the first found one and holding F2 pressed, it would add BPs to every call entrypoint on the search list.  So my impl. is not identical but achieves same effect.

I did a "find in files" for "Log String" and it didn't come up with anything of interest and I don't need the parameters if I have proper BP's (first set BP to everything in a file with a toggle, then run and watch which BP's get hit, then use toggle again to remove them and add manual BP to points of interest you discovered and then watch parameters or whatever).

bfiete:
I mean the Log String from Configure Breakpoint:
image

With that configured you can get a text output method calls with params without having to stop on each breakpoint.

But, yah- you probably want an option to set breakpoints at the start of all methods that are called from _outside_ the project they are defined in.

Either way, I'm not feeling like the particular functionality exposed by this PR is generally useful enough to be justified...

PR #1375, opened 2022-01-11
Highlight matching parens under cursor

blat-blatnik:
I often rely on this in other editors, especially in complex math expressions.

I have it enabled by default in all SourceEditWidgets, including the Immediate panel for example. It's also enabled by default and there is no option to turn it off. IMO I don't see why you would turn this off, but I can add it if you think I should. I did add a theming option for it so themes can change the color.

Untitled

### DIFF
--- a/BeefLibs/Beefy2D/src/theme/dark/DarkTheme.bf
+++ b/BeefLibs/Beefy2D/src/theme/dark/DarkTheme.bf
@@ -185,13 +185,14 @@ namespace Beefy.theme.dark
             COUNT
         };
 
-		public static uint32 COLOR_TEXT                = 0xFFFFFFFF;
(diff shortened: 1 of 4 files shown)


bfiete:
In your implementation, you checked to the right of the cursor before the left, and the comments mentioned that VS also does it that way. 

In my testing, Visual C++ _only_ attempts to the right, whereas Visual C# checks to the left for a closing character and also checks to the right for an opening character (it can do two hilite pairs at the same time if your cursor is between a `)` and a `(`, actually).

I modified the behavior to act like C#, minus supporting the two simultaneous pairs.

blat-blatnik:
Oh you're right about it checking only on the right. And I didn't realize that VS would treat c++ and c# differently with this one, but it makes sense to behave like VS c#.

PR #1400, opened 2022-01-28
fix save unnamed new file crash

EinScott:
when using File->New->New File, fix crash on save. Provide a reasonable default path for this rather rare case (everything else seems to already check for fielPath to be null). Maybe now viable option to create files in safe mode or just for non .bf files.


bfiete:
This PR crashes with an empty workspace. I committed a different fix at edcf92a451e14817945eb754f6cfab30ddf2e33b

PR #1415, opened 2022-02-06
Fixes to upgrade LLVM version to 13.0.1 

disarray2077:
First of all, I would like to clarify that this PR is still incomplete because of some things that I'm not sure how to fix, but even so, I have absolutely no idea if I missed something important, I have no experience with LLVM, and the only things that I fixed were the errors/crashes that I encountered, so I don't know if there is something else that needs to be solved.
Anyway, this PR does the necessary changes to compile Beef with LLVM 13.0.1 (latest version).

But, as I said, some things still need to be addressed:
- `setInlineAsmDiagnosticHandler` doesn't exist anymore, probably `setDiagnosticHandlerCallback` should be used instead, for now I just commented it out.
- The attribute `StructRet` changed from an `EnumAttr` to a `TypeAttr`, which causes LLVM to complain about it not being an `EnumAttr`, I'm not sure how to solve this but *maybe* we just need to use `getWithStructRetType`, for now I just commented it out.
- The attribute `ByVal` causes an assertion failure in LLVM: `Adding integer/type attribute without an argument.`, for now I just commented it out.


bfiete:
Hm- the ByVal isn't optional - that's required to match C calling conventions on non-Windows platforms ...

disarray2077:
I guess so, that's why I marked this PR as a draft until this (and probably the `StructRet` issue) can be solved.
Maybe the `setInlineAsmDiagnosticHandler` issue doesn't need to be solved since Beef currently doesn't support inline asm anyway.

bfiete:
Also- this doesn't include things like .vcxproj changes (those still point to the 11.0.0 release), or llvm_build.bat changes... I assume you also had to do those but they aren't part of this PR yet?

disarray2077:
I'm actually just using the 11.0.0 names 😅, I will do these changes soon

bfiete:
I'm also putting wasm fixes in this branch...

disarray2077:
I will remove the status of draft from this PR since StructRet/ByVal were fixed and the 11.0.0 > 13.0.1 changes were done, so I guess this is now ready to be merged, although wasm still needs some work, these are some problematic things that I found:
- The IDE doesn't have the `EmscriptenPath` setting in the UI
- `build_wasm` should also build BeefRT for wasm64
- `build_wasm` should output to `IDE\dist\Beef042RT[32/64]_wasm.lib` and the hardcoded `BeefRT.a` should be removed from the command line
- Beef doesn't add `-s MEMORY64` to the command line of `emcc` in wasm64 builds
- Beef should probably add `-g -s ASSERTIONS=1` for wasm debug builds
- wasm64 currently doesn't support pthreads, so we should disable threads in wasm64
- wasm32 should also be allowed to be compiled with threads disabled since thread support requires custom HTTP headers that aren't available everywhere (https://itch.io/t/1332996/cross-origin-isolation-to-unblock-wasm-sharedarraybuffers)
- wasm64 (with USE_PTHREADS=0) is failing to build with this error: `invalid memory limits flags`
- `emcc` isn't being executed when no object is generated, so if you only add a linker flag in the project it will not work
- `emcc` is being executed with the working directory set to `IDE/dist/llvm/bin`, probably the working directory should be the project directory for you to be able to use the `--embed-file` command-line option.
- wasm doesn't run in the browser with the default build type (Static Lib)
- wasm only compiles with GNU toolset

bfiete:
Good summary- thanks for the great work on this. I'm going to merge this asap.

Green-Sky (commented):
sorry if I come across like an *hole. :smile:

bfiete:
Did those actually cause issues or are you just being pedantic?

Green-Sky:
no I am just being pedantic.
I just thought it was funny that STRIP_TRAILING_WHITESPACE was now trailed by random whitespace...

PR #1452, opened 2022-02-15
Improve `Float.Parse`

disarray2077:
This PR does the following things to `Float.Parse`:
- Adds an optional `NumberFormatInfo` parameter
- Adds support for the `+` sign (which does nothing).
- Adds support for `NaN`, `Infinity`, `-Infinity`

### DIFF
--- a/BeefLibs/corlib/src/Float.bf
+++ b/BeefLibs/corlib/src/Float.bf
@@ -166,12 +166,29 @@ namespace System
 
 		public static Result<float> Parse(StringView val)
 		{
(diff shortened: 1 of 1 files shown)


bfiete:
I did a roundtrip test for all float values - iterating through all 4 billion values and passing them through `ToString` and then passing that string through `Parse` to ensure we get the same result back. It turned up a couple other things to fix but now it's solid.

Unfortunately testing all possible double values will take about 1.5 million years :-(

PR #1463, opened 2022-02-21
Add `LinkedList<T>` to corlib

disarray2077:
This is basically just an adaptation of C#'s code, but, unlike C#, you can't own the `LinkedListNode` because it can only be deleted when it's removed from a `LinkedList`. This looks sane for me but feedback is always welcome.


bfiete:
Yeah, this looks good!

PR #1480, opened 2022-03-01
Fix `Dictionary<T>.ValueEnumerator.Key`

disarray2077:
I also noticed that `ValueEnumerator` makes a copy of the value and `CurrentRef` is a ref to this copied value... This doesn't make much sense to me, so maybe we should change this too.


bfiete:
That's funny- I just noticed the same `CurrentRef` issue and I have a commit to fix it that I haven't pushed yet. I'll push in the next day or two... it's part of a pretty massive collection of changes.

Issue #1520, opened 2022-04-13
va_list is not treated properly on aarch64


hunterbridges:
Currently, `va_list` is treated as `void*` for interop purposes. However, on `aarch64`, `va_list` is a struct and has to be treated as such.

The [calling convention docs](https://c9x.me/compile/bib/abi-arm64.pdf) describe the layout as
```C
struct __va_list {
    void *__stack;
    void *__gr_top;
    void *__vr_top;
    int __gr_offs;
    int __vr_offs;
}
```

---

I did attempt a workaround which almost worked:

Beef calling frame:
```bf
[System.InlineAttribute] public static void Print(char8* destChars, uint32 maxLength, char8* format, ...)
{
	System.VarArgs vArgs = System.VarArgs();

	vArgs.Start!();
	Print_VA(destChars, maxLength, format, vArgs.ToVAList());
	vArgs.End!();
}
```

C++ receiving frame:
```cpp
void String::Print_VA(char8 *destChars, uint32 maxLength, const char8 *format, void *argPtr)
{
	va_list args = *(va_list*)argPtr;
	stbsp_vsprintf(destChars, format, args);
}
```

Dereferencing `argPtr` did yield a valid `va_list` struct, but the `format` pointer was invalid, and instead was filled with the binary values for the `__gr_offs` and `__vr_offs` fields. So it seems that calling convention is incorrect, and `va_list*` lining up with the valid data was merely coincidental

bfiete:
Out of curiousity-

if in VarArgs you change `int[3] mVAListl` to `int[5] mVAList;`, does that fix this workaround example?

hunterbridges:
Ah ok, yes, that does fix the workaround on Switch.

However, now there is an issue when running on Windows due to different levels of indirection used in `ToVAList` whether you're on Windows or not. But, if I also change that function to simply this:

```bf
public void* ToVAList() mut
{
	return &mVAList;
}
```

then I can use the same code in `Print_VA` above and get consistent results on both platforms

bfiete:
Just to check - does the `int[5]` change not fix the normal `ToVAList()` usage without the extra dereference hack?

It seems that it should.

hunterbridges:
It does fix it on aarch64. However the dereference hack is to make it so where consumers on the C++ side do not have to treat the incoming va_list pointer argument differently depending on the platform

hunterbridges:
[...]

PR #1525, opened 2022-04-16
VarArgs multiplatform compatibility tweak

hunterbridges:
Fix for #1520 

- Increased size of VarArgs.mVAList on non-Windows platforms
- Added ToVAListPtr() which returns &mVAList on all platforms

### DIFF
--- a/BeefLibs/corlib/src/Internal.bf
+++ b/BeefLibs/corlib/src/Internal.bf
@@ -66,7 +66,7 @@ namespace System
 #endif
 		}
 
-		public void** ToVAListPtr() mut
+		public void* ToVAListPtr() mut
 		{
 			return &mVAList;
 		}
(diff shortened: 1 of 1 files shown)


bfiete:
I don't think `ToVAListPtr` works on non-Windows machines... You'll get a conversion error from `int[5]*` to `void**`...

bfiete:
If it were just `void*` instead of `void**` then I think that'd be what you want. It's still not a "proper fix" but at least allows for the workaround you found.

hunterbridges:
oh my bad… I fixed that. I am not exactly sure what a proper fix would be though, since va_list is meant to be treated as a compiler built in type. I guess just digging in and implementing the various platform/compiler specific variants?

bfiete:
Yeah, the proper way is to have the actual return type of ToVAList match each platform's `va_list` in C. I think the majority are just pointers, though. I can attempt a proper fix for aarch64 at least, though.

Issue #1537, opened 2022-04-18
(aarch64) Wrong pointer passed to callee when getting pointer to struct member as a temporary


hunterbridges:
We have a class `WickPath` which extends `PathEntity`. `PathEntity` contains a member `Path path`, and `Path` is a struct defined as such:

```bf
[CRepr, Align(8)]
struct Path
{
    public PathType pathType;
    public uint8 editorFlags;
    uint32[7] opaque;
}
```

`WickPath` has a method `WickPath.Draw()` defined as such:

```bf
public override void Draw()
{
    Matrix matWorld = Matrix.Translate(path.GetPosition());
    Graphics.PushWorldMatrix(matWorld);

    Graphics.SetDrawLighting(true);
    DrawConnectingLinesPartial(&path, 0xFFFFFFCC, 4.0f, 0.0f, 1.0f - sparkProgress);
    DrawConnectingLinesPartial(&path, 0xFF111111, 4.0f, 1.0f - sparkProgress, 1.0f);

    Graphics.SetGPUState(.UI_DEPTH);
    Graphics.DrawBox(path.GetNodeAtIndex(path.GetNodeCount() - 1).position, HITBOX_SIZE, 0xFFFFFFCC, .ALL_CENTER);
    
    Graphics.PopWorldMatrix();
}
```

We were getting a crash in the first call to `DrawConnectingLinesPartial`. That method signature looks like this:
```bf
public static void DrawConnectingLinesPartial(Path* path, Color color, float lineThickness, float progressStart, float progressEnd)
```

When I stepped through, I noticed the address to `path` was still a valid pointer in heap space, but not the calling frame's `&this.path`. So it was a valid pointer, but not a valid `Path*`.

---

Then I tried taking the pointer and storing it in a local first like so:

```bf
Path* pathPtr = &path;
DrawConnectingLinesPartial(pathPtr, 0xFFFFFFCC, 4.0f, 0.0f, 1.0f - sparkProgress);

DrawConnectingLinesPartial(&path, 0xFFFFFFCC, 4.0f, 0.0f, 1.0f - sparkProgress);
```

The first call to `DrawConnectingLinesPartial`, using the `pathPtr` local, passed the correct pointer to the function. But the second call, taking the pointer into a temporary, passed the wrong pointer to the function.

Disassembly here:

```asm
			Path* pathPtr = &path;
0x082569A8:    add			 x9, x8, #208 ; 0xd0 
[...]

Issue #1521, opened 2022-04-13
Intermittent stack alignment faults on aarch64


hunterbridges:
Testing on Switch, we have run into intermittent cases where certain stack frames emit a stack alignment fault.

This seems like it may be related to [Stack Pointer Register usage rules](https://developer.arm.com/documentation/dui0801/a/Overview-of-AArch64-state/Stack-Pointer-register)

[This article](https://community.arm.com/arm-community-blogs/b/architectures-and-processors-blog/posts/using-the-stack-in-aarch32-and-aarch64) has a more in-depth break down of the stack usage rules on aarch64

---

We have noticed this occurring in an implicit conversion operator used as a temporary value in a method call to our C++ API. Beef Disassembly is here:

```asm
		public static implicit operator VertexStyle(Color color)
0x08370688:    sub			 sp, sp, #48 ; 0x30 ; RSDK::VertexStyle::operator RSDK.VertexStyle(RSDK::Color$part) 
0x0837068C:    str			 x30, [sp, #32] 
0x08370690:    stur			 x0, [sp, #28] 
			return VertexStyle(color);
0x08370694:    ldur			 x1, [sp, #28] 
0x08370698:    add			 x0, sp, #8 
0x0837069C:    bl			 0x8370514 ; RSDK::VertexStyle::__BfCtor(RSDK::Color$part) 
0x083706A0:    ldrh			 w13, [sp, #8] 
0x083706A4:    ldrh			 w12, [sp, #10] 
0x083706A8:    ldrh			 w11, [sp, #12] 
0x083706AC:    ldrh			 w10, [sp, #14] 
0x083706B0:    ldrb			 w9, [sp, #16] 
0x083706B4:    ldrb			 w8, [sp, #17] 
0x083706B8:    strh			 w13, [sp, #18] 
0x083706BC:    strh			 w12, [sp, #20] 
0x083706C0:    strh			 w11, [sp, #22] 
0x083706C4:    strh			 w10, [sp, #24] 
0x083706C8:    strb			 w9, [sp, #26] 
0x083706CC:    strb			 w8, [sp, #27] 
		}
0x083706D0:    ldrh			 w0, [sp, #18] 
0x083706D4:    ldrh			 w1, [sp, #20] 
0x083706D8:    ldrh			 w2, [sp, #22] 
0x083706DC:    ldrh			 w3, [sp, #24] 
0x083706E0:    ldrb			 w4, [sp, #26] 
[...]

PR #1538, opened 2022-04-18
Catch Int32 case in BfTypeInstance::GetLoweredType on Aarch64

hunterbridges:
Fix for #1537 (and also maybe #1521?)


hunterbridges:
Not a correct fix

PR #1540, opened 2022-04-20
[corelib- Stream] Return Ok when reaching the end of file

andreakarasho:
When reaching the end of the file the result of size is 0 and returns error.
Maybe it's necessary to apply this logic in other sections of the code.


bfiete:
This is in `Result<T> Read<T>()`
We must return an error here - there is no actual `T` value we can return!

bfiete:
And to be clear- this is not when _reaching_ the end of the file, this is when attempting to read _past_ the end of the file.

PR #1557, opened 2022-05-07
Print to Output window in IDE in Debug mode

wacom460:
This would allow you to see the Console output in a GUI application in the Beef IDE Output window only in debug mode.


bfiete:
I think the idea of allowing console output to be seen in the IDE is good, but I don't think this is the right way - we should have a separate Console tab, and it should show _all_ console output, not just console output from `Console.Write`.

I don't think this will be a trivial feature to implement.

PR #1565, opened 2022-05-17
added proper checking for double.Parse

Booklordofthedings:
it now loops through the string and errors if it find something that isnt a . or a digit
and counts the amount of .

other parsing functions would have to be redone in a similar matter later


bfiete:
I committed an error-handling version at 0ef11a89dc067e184fc3a153bd7152e92791c13b

Issue #1613, opened 2022-06-26
Struct that inherits enumerator can't be used in for loop


disarray2077:
The code below doesn't compile but I guess it should since `TestInheritEnumerator` inherits `StringSplitEnumerator`.
```cs
struct TestInheritEnumerator : StringSplitEnumerator
{
	public this(char8* ptr, int strLength, char8 splitChar, int count, StringSplitOptions splitOptions) : base(ptr, strLength, splitChar, count, splitOptions)
	{
	}

	public this(char8* ptr, int strLength, char8[] splitChars, int count, StringSplitOptions splitOptions) : base(ptr, strLength, splitChars, count, splitOptions)
	{
	}
}

public static void Main(String[] args)
{
	StringSplitEnumerator a = default;
	TestInheritEnumerator b = default;

	for (let s in a) // works
	{
	}

	for (let s in b) // must contain a 'GetEnumerator' method or implement an IEnumerator<T> interface
	{
	}
}
```
Tested with: https://github.com/beefytech/Beef/commit/52544e67821a6609c73ef676e3daabd7b8eba014

bfiete:
Fixed at 84bc0ad1de28532407d6af2c3c9ba1a045bdf083

Issue #1612, opened 2022-06-26
`extern` doesn't work inside generic type


disarray2077:
The code below doesn't compile, but it compiles if the type `TestExtern` isn't generic.

```cs
struct TestExtern<T>
{
	public extern void DoSomething();
}

extension TestExtern<T> where T : Char8
{
	public override void DoSomething() // No suitable method found to override
	{
		Console.WriteLine("Something");
	}
}
```
Tested with: 52544e67821a6609c73ef676e3daabd7b8eba014

bfiete:
Fixed at f2b177693d05ec27d2e9d88cddcfa45dab4d0047

PR #1614, opened 2022-06-27
Implement `String.Split` with StringView separators

disarray2077:
This PR does the following changes to the implementation of `String.Split`:
- When the separators argument is null or contains no characters, the split enumerator will treat white-space characters as the separators.
- `StringSplitEnumerator` has been refactored to `StringSplitEnumeratorBase<T>` to allow different implementations.
- `StringCharSplitEnumerator` and `StringStringSplitEnumerator` types have been added, they handle char8 and String splitting respectively.
- `String.Split` overloads for `StringView` have been added.

This PR depends on https://github.com/beefytech/Beef/issues/1613, since trying to compile any code with this change currently results in errors like the below:
```
ERROR: Type 'System.StringCharSplitEnumerator' must contain a 'GetEnumerator' method or implement an IEnumerator<T> interface at line 99:22 in d:\BeefLang\Repository\Beef\BeefLibs\SDL2\src\SDLApp.bf
    for (var line in contents.Split('\n', .RemoveEmptyEntries))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/BeefLibs/corlib/src/String.bf
+++ b/BeefLibs/corlib/src/String.bf
@@ -9,7 +9,6 @@ using System.Text;
 using System.Threading;
 using System.Interop;
 using System;
-using internal System;
 
 namespace System
 {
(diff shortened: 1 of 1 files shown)


bfiete:
The enumerators still don't work- you're basically trying to do virtual methods without virtual methods with `AdvancePos()` and `CheckMatch()`.

There _is_ a feature that allows extensions to statically override other methods from the original type declaration but that can't work with derivation with static dispatch because of the many-to-one relationship of derived types to a base type. 

A public 'MoveNext()' isn't actually needed - that's just legacy. You could make `MoveNext`  a static generic method and create an interface for Advance and CheckMatch and then the derived enumerators can just pass themselves into that generic `MoveNext`...

But personally I would just get rid of `StringSplitEnumeratorBase` and just repeat the code twice. I think it's justified here.

disarray2077:
Oh... I see, I guess the only thing that could possibly work still following this design would be what I tried at https://github.com/beefytech/Beef/issues/1612.
But anyway, my original code actually just had the code repeated twice but I thought it looked too rough and had too much repeated code so I tried to improve it xD, I guess I will just use it then.

PR #1630, opened 2022-07-06
Make `GetLocalTimeZoneFromTzFile` not crash and add willdcard check to `BfpFindFileData_CheckFilter`

disarray2077:
This PR removes the `NotImplemented` crash from `GetLocalTimeZoneFromTzFile` and adds the missing wildcard check to `BfpFindFileData_CheckFilter`.
About the algorithm used in the wildcard check, as explained in the source code it was obtained from the [wine-mirror repository](https://github.com/wine-mirror/wine/blob/master/dlls/ntdll/unix/file.c), I'm not sure if that could cause licensing issues though.


bfiete:
😢 

We can't use LGPL code here- this gets linked into users' code, which due to LGPL copyleft makes their code bound by LGPL. We can only use the more permissive licenses like MIT, BSD, and Apache...

bfiete:
I found another well-tested wildcard function we could steal at https://github.com/github/putty/blob/master/wildcard.c
The license is permissive.
Technically it's a superset of the Windows one, which gives the options of:
1) Remove the extra functionality
2) Let the Posix version have more capabilities
3) Add those capabilties to the Windows version as well

I think any of those options is acceptable, really.

disarray2077:
I did a bit more research and found one that fits us like a glove: https://github.com/FreeRDP/FreeRDP/blob/master/winpr/libwinpr/file/pattern.c
I guess it's also well tested, seems to match the Windows behavior, and is licensed under Apache :)

disarray2077:
> I found another well-tested wildcard function we could steal at https://github.com/github/putty/blob/master/wildcard.c The license is permissive. Technically it's a superset of the Windows one, which gives the options of:
> 
> 1. Remove the extra functionality
> 2. Let the Posix version have more capabilities
> 3. Add those capabilties to the Windows version as well
> 
> I think any of those options is acceptable, really.

Hm... If you say so then the more functionality the better 👍

disarray2077:
Done, Windows and Linux are now using the putty's wildcard algorithm.

disarray2077:
Ops, here is it 😅

Issue #1644, opened 2022-07-12
When running ./bin/build.sh on Mac I get an error that says 'ffi.h' file not found.


mileshenry31:
A screenshot of what my terminal looks like is attached. I already have libffi and cffi installed.

<img width="634" alt="Screen Shot 2022-07-11 at 9 40 46 PM" src="https://user-images.githubusercontent.com/73675082/178410239-25fc7fe2-43a4-4610-a0cc-ce02b8aea5a2.png">

mileshenry31:
On my Mac `ffi.h` is where you said it should be but in the Makefile the directory is `BeefySysLib\third_party\libffi\arm-apple-darwin-21.4.0/include`. Should I change it to `BeefySysLib\third_party\libffi\arm-apple-darwin-21.4.0`?

mileshenry31:
Fixed the error that said `ffi.h is missing` I explicitly told `Internal.cpp` where `ffi.h` was and changed `#include <ffitarget.h>` into `#include "ffitarget.h"` inside the `Internal.cpp` file.

disarray2077:
> On my Mac `ffi.h` is where you said it should be but in the Makefile the directory is `BeefySysLib\third_party\libffi\arm-apple-darwin-21.4.0/include`. Should I change it to `BeefySysLib\third_party\libffi\arm-apple-darwin-21.4.0`?

What... The abs_builddir is `BeefySysLib\third_party\libffi\arm-apple-darwin-21.4.0/include`? I didn't see this coming...

PR #1646, opened 2022-07-12
 Get libffi path from it's MakeFile

disarray2077:
Fixes https://github.com/beefytech/Beef/issues/1644

This PR hopefully should fix "libffi.h" include issues by getting the correct host target (which is necessary to find libffi.h) from llvm-config.

### DIFF
--- a/BeefRT/CMakeLists.txt
+++ b/BeefRT/CMakeLists.txt
@@ -45,20 +45,9 @@ if (HAVE_BACKTRACE_HEADERS)
    add_definitions(-DBFP_HAS_BACKTRACE)
 endif ()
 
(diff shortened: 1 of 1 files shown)


bfiete:
Did you test this already on Linux + Mac?

disarray2077:
I don't have a Mac :(, so I only tested it on Linux (x86_64). But I'm pretty sure it should work everywhere else 🤞

disarray2077:
`config.log` for some reason didn't exist in my other machine (which is also Linux x86_64), so I changed the code to just pick the absolute path from libffi's Makefile, I guess there's no way this can go wrong now.

PR #1681, opened 2022-08-13
Added bookmark panel

aharabada:
In short, this PR adds a panel to manage bookmarks.

Here is a list of things I implemented:
- Bookmarks panel
  - Shows a list of all bookmarks
  - Rename bookmarks
  - Change order of bookmarks via drag'n'drop (also affects order when jumping to next/prev bookmark)
  - Delete bookmarks
  - Put bookmarks into folders (folders cannot be nested)
- Cycle through all bookmarks or only the bookmarks in the current folder (can also be done with key-bindings, currently not bound)
- Enable/Disable bookmarks (disabled bookmarks will be skipped when jumping to next/prev bookmark)
- Bookmarks of the current workspace will be cleared when loading a different workspace
- IconButton received some love

Note: Since I changed how bookmarks are stored, all existing bookmarks will be lost

Screenshot:
bmp


bfiete:
I’m on vacation for the next week - looking forward to checking this out!

bfiete:
I made a couple tweaks at https://github.com/beefytech/Beef/commit/ebcaffbae9149377421431f2ec3cd296833128f3

Including auto-generating a bookmark name based on the syntactic source file location

PR #1696, opened 2022-08-26
Require listen IP when calling Socket.Listen()

m910q:
Current `Socket.Listen()` always listens on 127.0.0.1.
Now it is a required input on the method.

### DIFF
--- a/BeefLibs/corlib/src/Net/Socket.bf
+++ b/BeefLibs/corlib/src/Net/Socket.bf
@@ -117,7 +117,7 @@ namespace System.Net
 #endif
 
 		[CRepr]
-		struct in_addr
+		public struct IPv4Address
 		{
 			public uint8 b1;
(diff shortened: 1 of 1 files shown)


bfiete:
I think we should still have the version that defaults to `127.0.0.1` - whether or not it's a good default may be debatable, but removing that signature breaks backward compatibility.

m910q:
I had the same thoughts as you, but prioritized correctness over backwards compatibility.
I have changed it to default to 127.0.0.1.

Note that the new version still breaks compatibility in case the `backlog` parameter is set (without naming it). But the method starts getting weird of the address is the last parameter.

bfiete:
It might be best to just provide a second method with the original signature that calls the "full" implementation method...

m910q:
Alright, pushed.

PR #1730, opened 2022-10-05
Add linux filewatcher

Fusioon:
Implements filewatcher for linux using inotify API.
It should work exactly as filewatcher on Windows.


bfiete:
Thanks!

Fusioon:
Always pleasure contributing to this wonderful project! I made a test for filewatcher and noticed some differencies between between linux and windows (when using code to create/modify/delete because these events happen too fast) :

- When file is created (`File.WriteAllText`), modified and deleted, on linux it triggers, create + modify and delete, the one modify is lost
- File/directory creation doesn't trigger Modified on parent directory
- When creating directory and then file inside this directory, the file create+modify is lost because we haven't yet added watcher for this directory

Are these differences something that should be fixed?

bfiete:
The first one seems okay, the other two sound wrong.

PR #1756, opened 2022-11-21
Feature: Pinned tabs

kallisto56:
Hi there,
This feature adds the option to pin tabs to the left side of the tabbed view.
Some notes:
- This feature adds `bool mIsPinned` to `TabbedView.TabButton`;
- When `mUISettings.mInsertNewTabs` is set to `.RightOfExistingTabs`, new tabs will appear at the end, otherwise, any new tab will appear after the right-most pinned tab;
- This feature works with tabs like "Output" or "Call Stack", but those panels do not spawn context-menu, which means changing pinned state is only possible through drag-n-drop; (I'll probably figure out how to change that soon);
- Pin icon was drawn by me (meaning it is not copied from anywhere);
- Users will not notice the presence of this feature unless they specifically call context-menu action `Pin this tab`;

Feel free to change the code or reject the pull request entirely, this is your project, you are the boss.

https://user-images.githubusercontent.com/1311214/203152013-9ac591f8-647a-4854-9de8-15aeb9b4ea91.mp4


bfiete:
Looks good!

PR #1784, opened 2023-01-13
Add more constructors to `ObsoleteAttribute`

disarray2077:
This PR makes the `isError` parameter optional for the `ObsoleteAttribute` attribute.


bfiete:
Yeah, this is good. I originally thought 'isError' should be explicit, but the vast majority of the cases you just want it to be a warning.

PR #1785, opened 2023-01-25
added GetCursorPosition

Booklordofthedings:
Maybe it should return a Result<uint16[2]> instead but im not sure there, rn it just returns {0,0}

### DIFF
--- a/BeefLibs/corlib/BeefSpace.toml
+++ b/BeefLibs/corlib/BeefSpace.toml
@@ -1,5 +1,0 @@
-FileVersion = 1
-Projects = {corlib = {Path = "."}}
-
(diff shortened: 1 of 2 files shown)


bfiete:
It's probably best just to copy the C# properties here: https://learn.microsoft.com/en-us/dotnet/api/system.console?view=net-7.0

Booklordofthedings:
Should the properties throw errors ?
SetCursor fe. should error if the input value is too large, but I dont like the idea of making something unhandleable-

bfiete:
Clamping is probably the best behavior - your does not currently clamp, it just silently fails.

PR #1809, opened 2023-03-05
Fix probable typo in BfExprEvaluator.cpp

RogueMacro:


bfiete:
Does this fix a bug? If so, please post the issue that it addresses.

bfiete:
Null check fixed at 5c6eec125c6bfa343212c4090f9250b045c95d1e

PR #1823, opened 2023-04-01
Update SDL, SDL_image, SDL_mixer, SDL_ttf to newest release

daafu:
This change updates SDL & friends to the newest release versions.

The update strategy was to diff the C header files of existing Beef SDL versions to the headers of the current releases, and then integrate all new APIs/changes by hand. After this initial pass I went through all the APIs, enums, structs by hand to verify correctness.

Also fixed numerous issues with existing APIs. Vast majority of those issues were "int" vs "int32", and a few more serious issues like wrong APIs in LinkName, or structs with incorrect member types.

To test I compiled Beef, and also tested SpaceGame, as well as my own SDL project. Have not done extensive testing of new APIs.

This change may break some existing SDL projects in trivial ways. Eg, some methods still had "SDL_" prefix, which has been removed. Some ptr*s were changed to "out" variables.


bfiete:
Thanks! I'm not going to be able to check this out for a couple weeks, but I will get to it!

PR #1824, opened 2023-04-03
Simd improvements

MineGame159:


bfiete:
Thanks! I'm not going to be able to check this out for a couple weeks, but I will get to it!

PR #1837, opened 2023-05-03
Improve ILIst and add IDictionary

disarray2077:
This pull request introduces additional basic functionality to the existing IList interface and a new IDictionary interface with basic functionality.


bfiete:
Hm, the var extensions like `extension List<T> : IList where T : var` don't feel right...

I'm assuming you had to do that because there isn't an unconstrained `Variant.Get<T>` so you get an error. I think the right way to fix it would be to just add a `public T Get<T>() => default;` to Variant and then you should be able to get rid of those extensions...

disarray2077:
That makes sense, and that's right, I didn't even think about this option.

PR #1845, opened 2023-05-14
Improvements to Number parsing

disarray2077:
This PR makes some improvements to the Parse method of all Integer types:
- Normalizes all Parse methods.
- Adds an overflow check to all Parse methods.
- Modifies all Parse methods to accept Hex input only when the AllowHexSpecifier flag is set.
- Adds the Parse method for the types int8, int16, uint8, and uint16.

### DIFF
--- a/BeefLibs/corlib/src/Int.bf
+++ b/BeefLibs/corlib/src/Int.bf
@@ -9,11 +9,12 @@ namespace System
 		{
 			case Ok;
 			case NoValue;
+			case Overflow;
 			case InvalidChar(int partialResult);
 		}
 
-		public const int MaxValue = (sizeof(uint) == 8) ? 0x7FFFFFFFFFFFFFFFL : 0x7FFFFFFF;
-		public const int MinValue = (sizeof(uint) == 8) ? -0x8000000000000000L : -0x80000000;
+		public const int MaxValue = (sizeof(int) == 8) ? 0x7FFFFFFFFFFFFFFFL : 0x7FFFFFFF;
+		public const int MinValue = (sizeof(int) == 8) ? -0x8000000000000000L : -0x80000000;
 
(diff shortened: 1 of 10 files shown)


bfiete:
This did break some functionality with hex numbers.

Behavior table:
|Language|Input|Flags|Result|
|---|---|---|---|
|C#|"100"|AllowHexSpecifier|256|
|C#|"0x100"|AllowHexSpecifier|Error|
|C#|"100"|HexNumber|256|
|C#|"0x100"|HexNumber|Error|
|C#|"0x100"|none|Error|
|Old Beef|"100"|AllowHexSpecifier|256|
|Old Beef|"0x100"|AllowHexSpecifier|256|
|Old Beef|"100"|HexNumber|256|
|Old Beef|"0x100"|HexNumber|256|
|Old Beef|"0x100"|none|256|
|New Beef|"100"|AllowHexSpecifier|256|
|New Beef|"0x100"|AllowHexSpecifier|256|
|New Beef|"100"|HexNumber|256|
|New Beef|"0x100"|HexNumber|256|
|New Beef|"0x100"|none|Error|

See https://learn.microsoft.com/en-us/dotnet/api/system.globalization.numberstyles?view=net-7.0

The flag names in C# are very bad. `AllowHexSpecifier` actually *does not allow hex specifiers*. `HexNumber` just adds the `AllowLeadingWhite | AllowTrailingWhite` to the `AllowHexSpecifier` flag. There's no actual way to parse "0x" - the "official" way is to use `Convert.ToInt32`, which is dumb.

I think we should break with C# and make "AllowHexSpecifier" mean what it sounds like, and make "HexNumber" mean to always parse the number as hex...

bfiete:
Well, I did that: https://discord.com/channels/670260150656827422/670260151130914818/1121409642245931078

PR #1859, opened 2023-06-06
Update build_wasm.bat, remove -s WASM=1 and -s USE_PTHREADS=1

sere3925sere:
-s WASM=1 is outdated and does nothing nowadays.
-s USE_PTHREADS=1is replaced by -pthread flag, using -s USE_PTHREADS=1 causes issues (exception) when linked with libraries that are compiled without this flag. Might need to have 2 runtime libraries, one for singlethreaded programs and one for multithreaded ones. More info here https://emscripten.org/docs/porting/pthreads.html

### DIFF
--- a/IDE/src/IDEApp.bf
+++ b/IDE/src/IDEApp.bf
@@ -2479,7 +2479,10 @@ namespace IDE
 			ShowPanel(mOutputPanel, false);
 			mMainFrame.RehupSize();
 
-			ShowStartupFile();
+			if ((mSettings.mUISettings.mShowStartupPanel) && (!mWorkspace.IsInitialized))
+				ShowStartup();
+			else
+				ShowStartupFile();
 
 #if !CLI
 			if (mMainWindow != null)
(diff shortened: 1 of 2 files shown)


bfiete:
Oh- does the switch from `-s USE_PTHREADS=1` to `-pthread` fix your crash issue?

sere3925sere:
No, it still crashes. It works fine if neither -sUSE_PTHREADS=1 nor -pthread is present. Since my main program doesn't use threads, and library I use doesn't use threads, removing -pthread from runtime fixes crash. If I'll ever try to make a multithreaded program, I'll have a different issue, most likely.

bfiete:
Ah okay, I didn't actually look at the code but from the text I thought you were stating that you replaced `-s USE_PTHREADS=1` with `-pthread` but you did not actually add `-pthread`. That sortof just switches to problem to breaking anyone who actually _does_ want to use threads...

sere3925sere:
I don't know the codebase well enough to fix it properly. I think I need to:
*) Modify build_wasm.bat to generate Beef042RT32_wasm_pthread.a in addition to Beef042RT32_wasm.a
*) Modify beef compiler to use Beef042RT32_wasm_pthread.a instead of Beef042RT32_wasm.a depending on workspace/project options.
*) Run multithreading tests.
*) Try to link a library that uses multithreading, and see if that works.
I barely know this language, don't make me do this.

PR #1882, opened 2023-08-03
Add intrinsic for emscripten assembly/javascript.

hermansimensen:
Using emscripten there are three main ways of calling javascript code. 

The first one is the slowest, which is to use the emscripten_run_script export, which is essentially running all code through javascripts 'eval' function. This is not ideal if you are looking to get any sort of good performance.

The other two ways are through em_asm and em_js.

Both of these require that strings are put into their own data section in the binary called 'em_asm' and 'em_js'. If this is not the case, emscripten will not allow linking to the required exports for javascript code execution other than 'emscripten_run_script'. The commit attached to this PR adds a em_asm_internal intrinsic which adds strings to the em_asm data section of the binary, and returns a pointer to that string. 

You can read more about this here
[https://emscripten.org/docs/porting/connecting_cpp_and_javascript/Interacting-with-code.html#interacting-with-code-call-javascript-from-native](url)
[https://github.com/emscripten-core/emscripten/blob/main/system/include/emscripten/em_asm.h](url)
[https://github.com/emscripten-core/emscripten/blob/main/system/include/emscripten/em_js.h](url)

It would be used more or less in this way:

```beef
        [Intrinsic(":em_asm_internal")]
	public static extern char8* em_asm_internal(uint8* s);

	[CLink]
	public static extern void emscripten_asm_const_int(char8* code, char8* arg_sigs, ...);

	static Span<uint8> GetStringData(String value)
	{
	    return StringView(value).ToRawData();
	}

	static void em_asm<T>(T value, ...) where T : const String
	{
	    static uint8[?] data = GetStringData(value);

	    #unwarn
	    char8* str = em_asm_internal(&data);
	    emscripten_asm_const_int(str, "i");
	}

	public static void Main(String[] args)
	{
            em_asm("alert('hi');");
	}
```

There are a few things that could and maybe should be changed. 

The first thing being that we could rename em_asm_internal to something more generic and allow adding strings/data to any data section by specifying the data section as a second parameter. That way we could use the same intrinsic for both em_asm and em_js.

There also needs to be some helper code like seen in the example above. Should there be an emscripten library that contains all of emscriptens exports, or should this functionality be added to corelib somehow? 

For em_asm alone, there would at least need to be the importing of emscripten_asm_const_int as well as a call to the em_asm_internal intrinsic, as well as a call to figure out arg_sig for the emscripten_asm_const_int call, as seen in the em_asm header file linked to above. It might be cleaner to have all of it as its own library.

### DIFF
--- a/BeefLibs/corlib/src/WebAssembly.bf
+++ b/BeefLibs/corlib/src/WebAssembly.bf
@@ -1,0 +1,121 @@
+#if BF_PLATFORM_WASM
+
+namespace System;
+
+using System.Interop;
+
+class WebAssembly
+{
+	[CLink]
+	private static extern int32 emscripten_asm_const_int(char8* code, char8* arg_sigs, ...);
+	[CLink]
+	private static extern void emscripten_asm_const_ptr(char8* code, char8* arg_sigs, ...);
+	[CLink]
+	private static extern double emscripten_asm_const_double(char8* code, char8* arg_sigs, ...);
+
+	[Intrinsic(":add_string_to_section")]
+	private static extern char8* add_string_to_section(uint8* string, uint8* section);
+
+	private static Span<uint8> GetStringData(String value)
+	{
+	    return StringView(value).ToRawData();
+	}
+
+	/**
+	* Returns the string added to the specified data section. 
+	*/
+	private static char8* AddStringToSection<T0, T1>(T0 value, T1 section) where T0 : const String where T1 : const String
+	{
+		static uint8[?] data = GetStringData(value);
+		static uint8[?] sectionStr = GetStringData(section);
+
+	    #unwarn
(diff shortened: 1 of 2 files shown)


hermansimensen:
An example on how this now works:

```beef
	public static void Test()
	{
		int ret = JSCall<int...>("""
			console.log('I received: ' + $0 + ', ' + $1);
			return 56;
			""", (2, 5.1));

		JSCall<int...>("console.log('returned:' + $0)", ret);
	}
```

hermansimensen:
The em_asm stuff seems stable now. I haven't been able to get 'em_js' to work yet, so that might be a different PR. I get emscripten to recognize that there is content in the em_js data section, but it does not generate the code.

PR #1886, opened 2023-08-24
Add output window filtering options similar to what Visual Studio offers

daafu:
To use right click on output window, and select which messages should appear and which should be filtered out.

By default all types of messages appear as they normally would before this change.

These settings are persisted into workspace_user file

This change helps keep the clutter out of the output window so that program output logs are easier to read, especially on application startup.


bfiete:
Nice. I'm on vacation and it'll be a bit before I can look this over, but I will get to it asap. Thanks.

PR #1893, opened 2023-09-02
fix System.Compiler.Mixin taking const arguments

EW2010:


bfiete:
Can you explain this? `Compiler.Mixin` should work with any const string even when the callee is not a comptime method.

EW2010:
When should callee ever not be a comptime method. I mean if it isn't why not just write the code there without the mixin and if ur using const eval why is it making evaluating a code string and not just mixing it in?

bfiete:
Because the callsite might be referring to a constant string defined elsewhere - ie: another class, or generated from a different comptime method.

EW2010:
I've now moved the not const version into its own function.

bfiete:
Can you explain what the purpose of this PR is? IE: Examples of problems that it fixes or features that it enables?

bfiete:
I believe this was intended to fix this
```CS
[Comptime]
public static void CT<TCode>(TCode code) 
         where TCode : const String
{
    // Compilation error: "Non-constant argument for param text".
    Compiler.Mixin(code); // Compilation error.

    Compiler.Mixin(ConstRet(code)); // This works fine
}
```

That just appears to be a bug with generic const string params not being recognized as constants when passed to ConstEval methods (ie: `Compiler.Mixin`). `Compiler.Mixin` absolute _does_ require a const string, however.

EW2010:
Lets say you have a comptime method which atkes in code, computes it and then runs the computed version!

bfiete:
Just to be clear you understand what `Compiler.Mixin` does: it is equivalent to having typed in the code of the constant string at the `Compiler.Mixin` callsite location. This obviously only works if you have a constant string, and it doesn't have anything to do with comptime.

EW2010:
ok

Issue #1888, opened 2023-08-24
[Enhancement] rest/leftover function for StringSplitEnumerator


Booklordofthedings:
A function returning the rest of the stringView from a string split enumerator.
I think can implement and pr this myself, im just not sure wether it would fit for the corelib.
This could be used to consume parts of a string and then get everything left, and may be usefull for some parsing situations.
Whats your oppinion on including this in corlib ?

bfiete:
Yeah, that sounds fine. I'd probably have it be a property called `Remnant`...

PR #1905, opened 2023-11-27
added remnant property

Booklordofthedings:
The property returns a string view of what remains in the current string split enumerator.
The second commit adds a variant for StringStringSplitEnumerator and also fixes a potential offset error, for things like empty strings


Booklordofthedings:
#1888

bfiete:
I also added a `Separator` property too, for completeness.

PR #1909, opened 2023-12-04
Add new overloads for Break/SafeBreak

daafu:
These work like existing Break/SafeBreak but additionally take in a conditional parameter. This allows breaks like this:

if (x==1) {
    Debug.Break();
}

To turn into:

Debug.Break(x==1)


bfiete:
I'm not sure I get the utility... 

One of the core utilities of Assert is that it displays the expression and file location of the expression on failure. Also, "make sure this thing is always true" makes sense... but this doesn't feel like a very common usage case, and it's not really that much shorter than `if (x == 1) Debug.Break();`, and if you personally use it then it's easy enough to put it in your own libraries.

PR #1933, opened 2024-02-16
Allow method reference as the Sort comparer

disarray2077:
This PR allows the use of method references as the Sort comparer, from my tests this almost doubles the sorting speed (depending on the data being compared).

## Delegate
```cs
let rand = scope Random();

let nums = new List<int>();
defer delete nums;

nums.Resize(10000000);
for (int32 i = 0; i < nums.Count; i++)
{
	nums[i] = rand.NextI64();
}

let sw = Stopwatch.StartNew();
defer delete sw;

nums.Sort(scope (lhs, rhs) => lhs <=> rhs);

Console.WriteLine(sw.ElapsedMilliseconds); // 1208
Console.Read();
```

## Method Reference
```cs
let rand = scope Random();

let nums = new List<int>();
defer delete nums;

nums.Resize(10000000);
for (int32 i = 0; i < nums.Count; i++)
{
	nums[i] = rand.NextI64();
}

let sw = Stopwatch.StartNew();
defer delete sw;

nums.Sort((lhs, rhs) => lhs <=> rhs);

Console.WriteLine(sw.ElapsedMilliseconds); // 746
Console.Read();
```

### DIFF
--- a/BeefLibs/corlib/src/Array.bf
+++ b/BeefLibs/corlib/src/Array.bf
@@ -159,19 +159,40 @@ namespace System
 
 		public static void Sort<T>(T[] array, Comparison<T> comp)
 		{
-			var sorter = Sorter<T, void>(&array.[Friend]mFirstElement, null, array.[Friend]mLength, comp);
+			var sorter = Sorter<T, void, Comparison<T>>(&array.[Friend]mFirstElement, null, array.[Friend]mLength, comp);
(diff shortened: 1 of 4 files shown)


bfiete:
Ooh beautiful. Funny thing is that sorting is one of the original use cases I had in mind for the methodref feature but I never actually implemented it on these common corlib cases. Thanks!

PR #1934, opened 2024-02-16
Use pre-built LLVM if available (Linux)

disarray2077:
It's common to see people complain about the build times for LLVM, so I'm submitting this pull request to simplify the compilation process of Beef on Linux.
The build scripts have been modified to support pre-built LLVM. However, if LLVM is not installed, or if it's not version 13, the scripts will default to using the LLVM compiled from the source.


bfiete:
Alright, we'll give it a go! LLVM 18.1.0 is just about to release and I think I'm going to update to that when it does... probably about time, eh?

PR #1943, opened 2024-03-18
Allow Enum.Parse with numbers again

disarray2077:
The ability to parse the numerical value of enums using Enum.Parse was inadvertently(?) removed in commit 47c9305. This PR reinstates this functionality in the EnumParser class.

### DIFF
--- a/BeefLibs/corlib/src/Enum.bf
+++ b/BeefLibs/corlib/src/Enum.bf
@@ -414,6 +414,43 @@ namespace System
 				}
 			}
 
+			bool canParseNum = typeof(T).FieldCount > 0 && !hadPayload;
+			String ident = canParseNum ? "\t\t" : "\t";
+
+			if (canParseNum)
+			{
+				code.Append("\tif (int64 val = int64.Parse(str))\n\t{\n\t\tswitch ((T)val)\n\t\t{\n");
+				
+				List<int64> cases = scope List<int64>();
+
+				int caseIdx = 0;
+				for (var fieldInfo in typeof(T).GetFields())
+				{
+					if (!fieldInfo.IsEnumCase)
+						continue;
+
+					if (var fieldTypeInst = fieldInfo.FieldType as TypeInstance)
+					{
+						Variant variant = fieldInfo.GetValue(null).Get();
+						int64 value = variant.Get<int64>();
+	
+						if (cases.Contains(value))
+							continue;
+
+						if ((fieldTypeInst.IsTuple) && (fieldTypeInst.FieldCount > 0))
+							Runtime.NotImplemented();
+
+						code.AppendF($"\t\tcase .{fieldInfo.Name}:\n\t\t\treturn .Ok(.{fieldInfo.Name});\n");
+
+						cases.Add(value);
+					}
+	
+					caseIdx++;
+				}
+
+				code.Append("\t\t}\n\t}\n\telse\n\t{\n");
+			}
+
 			int caseIdx = 0;
 			for (var fieldInfo in typeof(T).GetFields())
 			{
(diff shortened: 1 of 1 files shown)


bfiete:
Yes, good catch!

bfiete:
Ah, but we are missing some important functionality here. It is nice that you are error-checking that the int matches a declared enum case, BUT, when enums are used as flags then a ToString of `.A|.B` would actually result in `"3"`, not `"A | B"`, and then that wouldn't round-trip back through your code properly.

Efficiently outputting `"A | B"` is difficult dynamically because you have to account for multi-bit cases and such, but I suppose that's another possible comptime-based feature.

bfiete:
I just punted and restored the old behavior by adding this before the `return .Err ;`...

```C#
if (int.Parse(str) case let .Ok(iVal))
    return .Ok((.)iVal);
```

bfiete:
You know, there is another really good optimization we could make to this for very large enums - binary search.

Instead of
```C#
if (str.Equals("Test_0", ignoreCase))
	return .Ok(.Test_0);
if (str.Equals("Test_1", ignoreCase))
	return .Ok(.Test_1);
if (str.Equals("Test_2", ignoreCase))
	return .Ok(.Test_2);
if (str.Equals("Test_3", ignoreCase))
	return .Ok(.Test_3);
if (str.Equals("Test_4", ignoreCase))
	return .Ok(.Test_4);
if (str.Equals("Test_5", ignoreCase))
	return .Ok(.Test_4);
if (str.Equals("Test_6", ignoreCase))
	return .Ok(.Test_4);
...
```

We would do

```C#
int comp7 = str.CompareTo("Test_7", ignoreCase);
if (comp7 < 0)
{
	int comp3 = str.CompareTo("Test_3", ignoreCase);
	if (comp3 < 0)
	{
		if (str.Equals("Test_0", ignoreCase))
			return .Ok(.Test_0);
		if (str.Equals("Test_1", ignoreCase))
			return .Ok(.Test_1);
		if (str.Equals("Test_2", ignoreCase))
			return .Ok(.Test_2);
	}
	if (comp3 == 0)
		return .Ok(.Test_3);
	if (str.Equals("Test_4", ignoreCase))
		return .Ok(.Test_4);
	if (str.Equals("Test_5", ignoreCase))
		return .Ok(.Test_4);
	if (str.Equals("Test_6", ignoreCase))
		return .Ok(.Test_4);
	if (int.Parse(str) case let .Ok(iVal))
		return .Ok((.)iVal);
}
```

disarray2077:
> Ah, but we are missing some important functionality here. It is nice that you are error-checking that the int matches a declared enum case, BUT, when enums are used as flags then a ToString of `.A|.B` would actually result in `"3"`, not `"A | B"`, and then that wouldn't round-trip back through your code properly.

Good catch, I didn't think about that 😅

>I just punted and restored the old behavior by adding this before the `return .Err ;`

I think a sad thing about doing it this way is that if the input is a number, you would have to fail every single comparison before reaching the number parsing... Although, I guess this overhead wouldn't be that big since the comparisons would fail fast.

>You know, there is another really good optimization we could make to this for very large enums - binary search.

True. And for this to provide a significant performance gain, the enum would indeed have to be quite large. In fact, depending on the specific case, like the example you gave where only the number at the end is different, using CompareTo could actually hurt performance instead of improving it.

bfiete:
We could dramatically improve number parsing speed with a `str[0].IsDigit` check at the start, since identifiers can never begin with a number.

And `String.CompareTo` requires the same number of comparisons as `String.Equals` - they both iterate until they find the first non-matching character. Even for small Enums (<10 elements) I think the String.Comparison would be (marginally) faster.

bfiete:
I realized there's actually a much better way. Instead of going from O(N) to O(log N), we go straight to O(1):

```C#
int hashCode = str.GetHashCode();
switch (hashCode % 5)
{
case 0:
	if ((hashCode == 79988194400) && (str == "Test_0"))
		return .Ok(.Test_0);
case 1:
	if ((hashCode == 79988194496) && (str == "Test_3"))
		return .Ok(.Test_3);
case 2:
	if ((hashCode == 79988194432) && (str == "Test_1"))
		return .Ok(.Test_1);
case 3:
	if ((hashCode == 79988194528) && (str == "Test_4"))
		return .Ok(.Test_4);
case 4:
	if ((hashCode == 79988194464) && (str == "Test_2"))
		return .Ok(.Test_2);
}
```

bfiete:
We avoid switching directly on `hashCode` in order to ensure the switch creates a straight-forward jump table...

disarray2077:
>I realized there's actually a much better way. Instead of going from O(N) to O(log N), we go straight to O(1)

This is quite nice... However, looks like this change increases my compile time from 3 seconds to 15 seconds 😓

disarray2077:
>I realized there's actually a much better way. Instead of going from O(N) to O(log N), we go straight to O(1)

Nice! However, this seems to have made things heavier for me when doing a clean build, the compile time went from about 6s to 20s, but I guess that shouldn't matter if this only happens with clean builds.

bfiete:
Yikes. That's unexpected... I'll see if I can reproduce.

bfiete:
How many elements do you have in your biggest enum?

bfiete:
A 1000-element enum took about 1 second on my machine with Og+. 200ms of that was comptime generation and the rest was actual compilation...

disarray2077:
>How many elements do you have in your biggest enum?

1284, I also tried to make a big enum in another project, but the comptime didn't take that long, hm... I'll try to find out where the time is spent.

bfiete:
At the end of the compile the total comptime time is reported. What does that say? That should tell us if the time is in the compiler or in the comptime generation...

disarray2077:
Comptime execution time: 12.26s
Execution time: 1.28s
Total build time: 20.19s

image

disarray2077:
Looks like this can be reproduced with the below enum:
https://pastebin.com/Kb7rDQca

bfiete:
Ah, I'll check that out. I was going to suggest you try inserting this into the EnumParser.OnTypeInit to see the individual times, but looks like that's not necessary:
```C#
var startTick = (int64)Platform.BfpSystem_GetTimeStamp();
defer
{
	if (Compiler.IsBuilding)
		Debug.WriteLine($"EnumParser {typeof(T)} {((int64)Platform.BfpSystem_GetTimeStamp() - startTick) / 10000}ms");
}
```

bfiete:
I should make the normal Stopwatch work with comptime.

bfiete:
Okay- the algorithm I was using to find a "good" hash modulo was the issue. That's fixed.
Also I switched to generating a static hash table instead of emitting a huge switch. The runtime speed should be nearly the as good, but compilation speed and binary size is much better now.

Now your big enum takes 34ms of comptime and 116ms of compile time on my machine.

disarray2077:
Can confirm, it's pretty smooth now 👍 
Thank you!

PR #1957, opened 2024-04-27
Add Mixin methog flag

zerkawei:
This PR adds a flag to methods' data in order to tell if a method is a mixin.


bfiete:
Looks good! Thanks.

PR #1982, opened 2024-05-17
Implement IRefEnumerator<T*> in HashSet Enumerator

disarray2077:
This PR implements the IRefEnumerator<T*> interface in the HashSet Enumerator.

I'm not sure if the absence of this enumerator was intentional, but I suppose it would make sense if you didn't want the user to be able to modify the value via the Enumerator. However, considering that the TryAdd method already permits altering the value after insertion into the HashSet, it might be worth reconsidering the necessity of omitting the enumerator.


bfiete:
It wasn't intentionally omitted.

PR #1996, opened 2024-07-16
Added an open in terminal dialogue

Booklordofthedings:
This pull requests adds an open in terminal option to all cases where an open in explorer option is available.
It also adds a Settings option to preferences (Default Console). That option currently defaults to powershell.


bfiete:
"Default Console" probably isn't a very good name for the option... "Powershell" is the *default*, but the actual option should probably be called "Windows Console".

bfiete:
Why leave the "old" field name? Backwards compatibility with 20 minutes ago for one person? ;-)

Booklordofthedings:
fix that

PR #2008, opened 2024-08-14
[Enhancement] Name field for test attributes

Booklordofthedings:
This pr adds the Name field to the test attribute.

When running tests it currently uses the Typename+Methodname for the name of the test, which can sometimes
look a bit weird with longer types also this means that methods needs to have a long descriptive name for better differentiation.

However now, when the Name field is set, it will overwrite the name of the test to whichever name the user gives it.
Allowing for shorter names and better visibility on which test does what.

The name is also escaped, to avoid writing newlines and such directly to the beef output panel.

### DIFF
--- a/BeefLibs/corlib/src/Attribute.bf
+++ b/BeefLibs/corlib/src/Attribute.bf
@@ -457,6 +457,7 @@ namespace System
 		public bool Ignore;
 		public bool Profile;
 		public String Tag;
+  		public String Name;
 	}
 
 	public struct ImportAttribute : Attribute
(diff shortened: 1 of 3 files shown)


bfiete:
What's "Ovwn" mean?

Booklordofthedings:
Overwritename
Did I leave that in the final result ?

bfiete:
Nope. you replaced it with "Name"
Btw- this leaks: `String* str = bfModule->GetStringPoolString(field.mParam.mValue, typeInstance->mConstHolder);`. I'm fixing, and using a proper string escaper.

PR #2024, opened 2024-09-06
Fix startup panel mem leak & resize when item is deleted

Fusioon:


bfiete:
Have you verified that `mRecentsParent.mParent?.RehupSize();` is an acceptable substitution for `mRecentsParent.RehupSize();`? Those are not equivalent in functionality. You sure we don't just need a `if (mRecentsParent.mParent != null)` check?

Fusioon:
Yes i'm sure it's what i want, the problem is that after the deletion of the recents item the other items are not repositioned and it ends up looking like this:
image

But your question made me think about it a bit more and i opted for another solution :)

PR #2027, opened 2024-09-08
Implement xdg-desktop-portal for Linux dialogs

zerkawei:
Implements the different file dialogs for Linux using desktop portals. It should work on most DEs and adds a dependency to libsystemd. Further work may be needed down the road to make the dialogs' public interface more OS agnostic or to integrate a library like [Native File Dialog Extended](https://github.com/btzy/nativefiledialog-extended). For now the Linux implementation follows the Windows interface to be compatible with existing code.


bfiete:
Nice! Thanks for all the work.

PR #2052, opened 2024-10-29
Add Directory.Copy to corlib

Booklordofthedings:
This solution is a relativly elegant one imo.

The it copies fromPath onto toPath, not into. While that isnt in line with how the gui does it, its also how other copy functionality works, so it seemed fine here.

Ive also added a comment that indicates that it does this and also that it overwrites files.

The current behaviour may not play nice with deep paths, but thats not really of concern for beefs usecases so if someone needs the security they could just implement a stack safe version themselves.

### DIFF
--- a/BeefLibs/corlib/src/IO/Directory.bf
+++ b/BeefLibs/corlib/src/IO/Directory.bf
@@ -90,6 +90,23 @@ namespace System.IO
 			return .Ok;
 		}
 
+  		///Copies and overwrites the contents of fromPath into toPath.
+		public static Result<void, Platform.BfpFileResult> Copy(StringView fromPath, StringView toPath)
+		{
+			if(Directory.CreateDirectory(toPath) case .Err(let err))
+				return .Err(err);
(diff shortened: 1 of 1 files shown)


bfiete:
Btw- I did a small refactor
```C#
public static Result<void, Platform.BfpFileResult> Copy(StringView fromPath, StringView toPath)
{
	Try!(Directory.CreateDirectory(toPath));
	for (var file in Directory.EnumerateFiles(fromPath))
		Try!(File.Copy(file.GetFilePath(.. scope .()), file.GetFileName(.. scope $"{toPath}/")));
	for (var dir in Directory.EnumerateDirectories(fromPath))
		Try!(Directory.Copy(dir.GetFilePath(.. scope .()), dir.GetFileName(.. scope $"{toPath}/")));
	return .Ok;
}
```

- Removed unnecessary string allocations
- Used Try! mixin
- Coding style change to match corlib style ie: space in `for (`

PR #2074, opened 2024-12-01
Add CompactList<T> constructor when T is a class

m910q:
This previously worked without an explicit constructor, but now it's needed.


bfiete:
This was actually just a bug. Fixded at b900477287562ce68e2a6570038cf93f590efd11

PR #2111, opened 2025-01-25
Fix: Scrolling inside DarkListView and DarkMenuContainer based on Font.GetLineSpacing()

kallisto56:
Hi there,
This pull-request fixes an issue with scrolling inside some of the panels.

`class ProjectPanel` contains vertical scrollbar with `mScrollIncrement` set to 0. In this case, scrolling amount will be equal to `mPageSize / 10.0f` calculated by `Scrollbar.GetScrollIncrement()`. If workspace contains a lot of files, scrolling become a bit unpredictable. This has been fixed in `DarkMenuWidget` as you requested.

Same behaviour was in `class NavigationBar`, which is fixed in `DarkMenuContainer`.


bfiete:
Kinda curious if anyone is going to complain about the mouse scrolling being "too slow" in those areas now!

This matches behavior of most other apps with list views now though.

PR #2114, opened 2025-01-28
Remove some hardcoded colors to make the IDE more friendly to light themes.

MineBill:
During the day i work in a very bright room and i prefer using light colors until the evening. In an attempt to create a light theme, i discovered several places where text colors were hardcoded to white, making it impossible to read that text in a light theme. To help with this, i changed most of these to use the `DarkTheme.COLOR_TEXT` color and i had to introduce 4 new theme options and 2 new colors on the `DarkTheme` struct.

In every single change, i made sure the default dark theme stayed the same by using the previously hardcoded colors as defaults for the new settings/color options.

I'm sure i missed a few spots here and there but i prioritized immediate problems and if this PR is accepted, i could open another one changing what's left over.

*I noticed that the word 'hilite' was used instead of 'highlight' in multiple places, so i assume it is a deliberate decision but i didn't use it for the 'mCodeHighlight' and 'mCodeHighlightUnfocused' options. Please let me know if you want me to change the names.*

### DIFF
--- a/BeefLibs/Beefy2D/src/theme/dark/DarkCheckBox.bf
+++ b/BeefLibs/Beefy2D/src/theme/dark/DarkCheckBox.bf
@@ -121,7 +121,8 @@ namespace Beefy.theme.dark
             {
 				g.SetFont(mFont);
 
-				DarkTheme.DrawUnderlined(g, mLabel, GS!(22), GS!(-1));
+				using(g.PushColor(DarkTheme.COLOR_TEXT))
+					DarkTheme.DrawUnderlined(g, mLabel, GS!(22), GS!(-1));
 
 				/*int underlinePos = mLabel.IndexOf('&');
 				if ((underlinePos != -1) && (underlinePos < mLabel.Length - 1))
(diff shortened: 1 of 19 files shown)


bfiete:
Yeah, we spell it "hilite" here. This is literally in the PR right now which seems kinda silly having these two lines back-to-back:
```C#
GetColor("CharPairHilite", ref mCharPairHilite);
GetColor("CodeHighlight", ref mCodeHighlight);
```

MineBill:
Alright, i renamed them and fixed a few other dialogs and panels i noticed. I'm pretty sure i got most of them. There are few places where it could do with some extra colors (like the diagnostics panel) but at least almost everything is legible now when using light colors.

bfiete:
This looks good. Do you have your theme up on GitHub? It could be useful for others!

MineBill:
Not yet. Once it's done I'll add it [here](https://github.com/Jonathan-Racaud/awesome-beef?tab=readme-ov-file#themes) and post it in the Discord, so for anyone reading this in the future, look there.

PR #2136, opened 2025-02-14
Add a 'Line Height' options that allows changing the the line height of the text editor.

MineBill:
I found the default line spacing to be too "squished" for my taste, so i made the line height configurable, ensuring the default(1.0) doesn't affect existing users.

| Line Height=1.0 (default) | Line Height=2.0 |
| ---------------------- | ---------------------- |
| image | image |

Few things to note:
* Only the `SourceEditWidgetContent` uses the line height setting, other `DarkEditWidget`s use the default (1.0). 
* The cursor always uses the height of the font, instead of the line height, similar to JetBrains IDEs.
* Selection height and a few other things use the modified line height instead of the true line height, again following JetBrains IDEs because i'm more accustomed to them.

| Visual Studio | JetBrains Rider |
| --------------- | --------------- |
| image | image |

Some IDEs (JetBrains, VSCode) call their setting "Line Height" but Visual Studio calls it "Line spacing". I think both make sense but I chose "Line Height".

### DIFF
--- a/BeefLibs/Beefy2D/src/gfx/Font.bf
+++ b/BeefLibs/Beefy2D/src/gfx/Font.bf
@@ -894,11 +894,11 @@ namespace Beefy.gfx
             return (int32)theString.Length;
         }
 
-        public float GetLineSpacing(float heightMultiplier = 1.0f)
+        public float GetLineSpacing()
         {
 			if (mFTFont == null)
 				return 0;
-            return mFTFont.mHeight * heightMultiplier;
+            return mFTFont.mHeight;
         }
 
 		public float GetHeight()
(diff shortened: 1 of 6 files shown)


bfiete:
This doesn't really seem right to me:
```C#
public float GetLineSpacing(float heightMultiplier = 1.0f)
{
    if (mFTFont == null)
        return 0;
    return mFTFont.mHeight * heightMultiplier;
}
```

It seems kindof weird to pass a multiplier in there just do you can do `font.GetLineSpacing(lineHeight)` instead of `font.GetLineSpacing()*lineHeight`....

MineBill:
That should be all, unless you have extra feedback. I'm not sure what you meant by "nearest pixel", i assume it's just rounding the `GetLineSpacing()*lineHeight` multiplication? Also, do you want me to squash the commits into one to help with noise?

bfiete:
Yes, I meant rounding `GetLineSpacing()*lineHeight`. I don't need the commits in one, it's fine as it is.

bfiete:
I thought it looked a little weird to have `Line Height` as `1`. I thought this made more sense:
{D6CDC807-D152-4420-8A3B-ADD0E6360E92}

PR #2171, opened 2025-03-01
Fixed crash when defer scope target doesn't exist

aharabada:
This small change fixes a crash when the specified scope of a defer-statement couldn't be found.

Example Code that crashes without this fix:
```bf
{
	defer:MyScope Console.WriteLine();
	return 0;
}
```


bfiete:
Ooh, we don't get a lot of bug fix PRs. Thanks!

Issue #2176, opened 2025-03-04
IDE crash using static pointers


Grimelios:
```CSharp
namespace Scratch
{
    public struct Point : this(float x, float y)
    {
        // Crash!
        public static let ZERO_PTR = &ZERO;
        public static let ZERO = Self(0, 0);
    }
}
```

This code crashes the IDE on 03/03/2025. Once the crash occurs, it will continue to occur immediately when the IDE is re-launched.

The crash appears to relate to the `let` keyword, and more specifically when _both_ declarations use `let`. None of the following examples trigger a crash despite each expressing the same logic as the original:

```CSharp
public struct Point : this(float x, float y)
{
    // No crash!
    public static Self* ZERO_PTR = &ZERO;
    public static Self ZERO = .(0, 0);
}
```

```CSharp
public struct Point : this(float x, float y)
{
    // No crash!
    public static Self* ZERO_PTR = &ZERO;
    public static let ZERO = Self(0, 0);
}
```

```CSharp
public struct Point : this(float x, float y)
{
    // No crash!
    public static let ZERO_PTR = &ZERO;
    public static Self ZERO = .(0, 0);
}
```

PR #2177, opened 2025-03-06
Fix static readonly warning in class body

Fusioon:
Fixes: #2176
Just adds a null check.
While toying with this i have noticed that readonly fields are marked as modifiable in ctors / init blocks, so it won't generate an warning when used as in the issue.

Also found an another issue:
```
class Program
{
	struct Point : this(int32 x, int32 y)
	{
		public static let sValue = Point(10, 10);
	}

	static int Main(System.String[] args)
	{
		let q = Point.sValue;
		return 0;
	}
}
```
This code is enough to trigger an crash on compile. 

I have tried to fix it, but tbh i have no idea if it's even remotely right, here is link to the branch: https://github.com/Fusioon/Beef/commit/0fe47816644c0c4234b1a5ef4fbbe791cb2a7895


bfiete:
Yay, bug fix PR!

And the FinishInit fix has been committed at ff7a9896cd1ca7b4a3d5d0a42b07a7679a804994

PR #2181, opened 2025-03-09
Fix enum underlying const eval on static fields

Fusioon:
Using `.Underlying` on static fields makes the compiler error out with `Non-constant argument for param 'this'`
I don't think it should be const evaluated so here is a fix

Code to better visualize what i'm talking about:
```
enum TestEnum
{
	Zero,
	One
}

static TestEnum Value = .One;
static void Example(int i) => Console.WriteLine(i);

public static void Main(String[] args)
{
	Example(TestEnum.Zero.Underlying);
	Example(Value.Underlying); // currently an error, fixed by this PR
}
```


bfiete:
I had a slightly different fix here: https://github.com/beefytech/Beef/commit/ba558394f1530206c246ea8d853f1585a1367595

There's a 'gotcha' in the const system where global variables are technically a kind of 'const' even those their values are not actually constant.

Issue #2164, opened 2025-02-24
[Bug][Crash] Failure during codegen for a switch that uses variable from a static field type of enum with payload


kallisto56:
Hi there,

Note: this issue is not a regression, it is reproducible on older builds.
This issue boils down to using a switch for an enum with payload. It must be static field (global or another class), otherwise failure does not happen.
Using enum with payload in if-statements works fine.
Reproducible on latest nightly (https://github.com/beefytech/Beef/commit/1805316b70387a8cfabdee45f1e5204fa0c5d649), as well as previous builds.
Ready-for-testing project: [bug.2164.zip](https://github.com/user-attachments/files/18936950/bug.2164.zip)
This issue is related to https://github.com/beefytech/Beef/issues/1780 but focuses solely on the failure during codegen. Apologies for creating a duplicate, I forgot that in 1780 it was mentioned that it crashes...
```C#
class Program
{
	static Food sFood;

	static void Main ()
	{
		switch (sFood)
		{
		case .Beef(let cookedLevel): Debug.WriteLine("Beef({})", cookedLevel);
		default: Debug.WriteLine("...");
		}
	}
}

enum Food
{
	case Beef (float cookedLevel);
}
```
Message box:
```
FATAL ERROR
Failure during codegen of ?Main@Program@bug1780@bf@@CAXXZ in bug1780_Program:
Unable to find bevalue for operand: %switch.0
 MCBlock: caseMatch:4
```

Issue #1780, opened 2023-01-01
static payload enum variable behaves different in switch than local one


EinScott:
```
enum A
{
	case A(int a);
	case B;
}

static A a;

static void Main()
{
	//let a = a;
	switch (a)
	{
	case .A(let intThing):
	case .B:
	// default:
	}
}
```

Without the commented-out local variable it demands a default case, and even if you give it that, compiling compiling crashes in code-gen.

With the local variable there, it accepts the switch without default and is fine compiling it.

PR #2185, opened 2025-03-12
Fix static payload enum switch isConst

Fusioon:
Fixes: #2164, #1780

But there is still a slight problem I'm unsure how to fix, `mBfIRBuilder->IsConstValue` does not work reliably when we are not generating code eg. (autocomplete/mouse over), i believe the ids are not valid in the context of the IRConstHolder, same thing i noticed with https://github.com/beefytech/Beef/commit/ba558394f1530206c246ea8d853f1585a1367595 the `.Underlying` still gets red squiggles when i mouse over but does compile fine. I was able prevent this by adding `(!mBfIRBuilder->mIgnoreWrites)` before checking for const be that seems like a huge hack.


bfiete:
The mention you issued is now fixed at 11ccb876a3685f66d2c5e945045b156bd59530a2

The issue was how we were "faking" static field references during autocomplete.

PR #2195, opened 2025-03-16
Fixed generated methods for enums with CRepr attribute

aharabada:
The recent changes to `CRepr` which make it so that `this` is always passed to methods as a pointer, broke the methods `HasFlag`, `GetUnderlying`, `GetUnderlyingRef` and `ToString` of enums that have the `CRepr`-Attribute. The compiler crashes during code gen because it receives a value of pointer type where it expected an integer value.

This PR fixes these issues by emitting load instructions (if necessary) while generating the IR for the methods in BfModule.
It also adds a test case. Without the changes in this PR the compiler crashes while compiling the test.


bfiete:
Oh wow a bug fix with a test! I love it!

PR #2202, opened 2025-03-28
Add IPv6 support to Socket

zerkawei:
This PR adds IPv6 to `System.Net.Socket` by :

- Adding the `IPv6Address` and `SockAddr_in6` structs
- Adding new `Listen` , `RecvFrom`, `SendTo` overload with IPv6
- Adding a new `Connect` overload that uses `getaddrinfo` instead of the deprecated `gethostbyname`
- Making `AcceptFrom` like `RecvFrom` and `SendTo`
- Adding `OpenUDPIPv6` (Open to better name suggestions)

### DIFF
--- a/BeefLibs/corlib/src/Net/Socket.bf
+++ b/BeefLibs/corlib/src/Net/Socket.bf
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics;
+using System.Interop;
 
 namespace System.Net
 {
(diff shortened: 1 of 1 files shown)


bfiete:
Good work!

One thing that I don't think really works is the overloading on out types. Code that would have previously worked with `socket.AcceptFrom(listenSocket, var clientAddr);` will now generate an ambiguous method selection error.

zerkawei:
Good point, I hadn't thought of this case. I'll remove the overloads on out types for now as they were purely a convenience and the methods with SockAddr* can do the work just fine.

PR #2203, opened 2025-03-28
Custom Build Properties: Proof of Concept Implementation

es-yuvanw:
This PR probably needs more refinement to be production-ready, but I just wanted to get a conversation going in case this would be a useful feature.

### Background:

Beef currently allows users to specify certain built-in properties like "$(ProjectName)", "$(WorkspaceDir)", etc. in the BeefProj.toml config file, which are resolved when running the IDE or BeefBuild. This system has a few limitations at the moment:

- Build properties can only be specified in BeefProj.toml and not BeefSpace.toml.
- Build properties can only be used as values for certain items in BeefProj.toml (for example, you can’t use a property as the value for “Platform.Windows.ProductVersion”).
- Users are limited to the current set of built-in properties and can’t define their own.

### Feature Request:

Ideally, we’d like to extend the current build property system to have the following functionality:

- Users can define custom build properties.
- Users can override custom build property values via the command line when running BeefBuild.
- Properties can be specified in both BeefProj.toml and BeefSpace.toml
- Properties can be used in the values for any item in BeefProj.toml and BeefSpace.toml

For additional context, our concrete use cases for this feature are:

- Specifying the Window’s product version for Beef apps via a build property, and being able to override it via the command line when creating new builds.
- The install path for our game engine may differ for each member of our team, so we’d like to avoid hardcoded paths to the engine location in BeefProj.toml and BeefSpace.toml. Instead, we’d like to specify the engine install path as a custom build property so the contents of BeefProj.toml and BeefSpace.toml can remain the same and still work for the whole team.
- Specifying the game engine’s version as a custom build property so that migrating apps to newer engine versions is easier (you just need to update the value of the engine version’s build property).

### Implementation Proposal:

I’ve implemented a proof of concept version of custom build properties in this PR. Custom build properties are specified in a new config file called BeefProperties.toml, which only contains definitions for custom build properties:

BeefProperties.toml
```toml
[Properties]
EngineInstallPath="C:/Program Files/Engine"
EngineVersion="1.2.3.4"
AppVersion="5.6.7.8"
```

I decided to create a new config file so custom build properties can be referenced by both BeefProj.toml and BeefSpace.toml (though I suppose placing them in BeefSpace.toml could work as well). There is also a new “-property” command line option in BeefBuild for specifying or overriding the value of custom build properties.

```
BeefBuild.exe -property="AppVersion=9.9.9.9"
```

Most of the implementation is contained in a new file called `CustomBuildProperties.bf` to make updating/reapplying changes easier in case we end up needing to maintain our own Beef distro. Here are the broad strokes of the implementation:

- Properties are loaded from BeefProperties.toml (if available) whenever a workspace is loaded.
- Upon receiving a request to compile a project, any data members of the project that contain custom build properties get resolved (e.g. “$(EngineVersion)” is replaced with “1.2.3.4”). After the project is queued for compilation, the project’s data members get unresolved to the original data (e.g. “1.2.3.4” reverts back to “$(EngineVersion)”). The unresolving process is done so that 1) when viewing project settings in the IDE, the custom build properties still show up as variables, and 2) if the project settings are saved/serialized to BeefProj.toml through the IDE, the custom build properties don’t get replaced with their resolved values.
- For workspaces, this implementation only supports custom build properties in the “Projects” item of BeefSpace.toml at the moment, but I think a similar strategy to how custom build properties are resolved/unresolved in projects could be used.

### Known Issues:

- The resolve/unresolve logic for projects may be a bit overkill/inefficient at the moment. To make sure we can unresolve custom build properties in project data, a deep copy of the project is created before its properties get resolved, and then that copy is used to restore the unresolved properties. As an alternative, I could call ResolveConfigString() before any project/workspace String data member is accessed (which from my understanding is the current convention in the codebase for built-in properties), but I avoided that route because it would involve a lot more invasive changes and seemed bug-prone.
- I couldn’t seem to find a clean place to resolve/unresolve custom build properties for workspaces, because workspace data seems to be used at various points throughout the IDE lifetime, compared to project data which seems to mainly be used when building the project.


hunterbridges:
For some context @bfiete, we need this to accomplish some things inside of our build chain.

I have looked at @es-yuvanw's implementation myself, and I think one thing that would need to change is how the variables are being statefully resolved and unresolved. It should probably be more like, the values stored on the project are the "unresolved" or virtual values, then at compile time, a copy of those values are made and resolved, and those values are read internally by the compiler.

Or maybe they are read via an interface that can lazily resolve whichever ones are requested, and has an internal cache.

By having the property resolution be stateful as the implementation currently is, it is very likely the variables would unintentionally end up in the opposite state as intended because something went wrong.

es-yuvanw:
Hi @bfiete, thanks for taking the time to review this PR! Just wanted to check in and see if you had any further feedback on this. We’d love to be able to use this feature in our build chain, and I’d be happy to modify the implementation as needed.

bfiete:
I think the blocker for adoption right now is the "temporary member resolution" technique - I would prefer for Project/Workspace member values to not change during compilation. I believe an effective approach could be just to insert some "properly resolution" calls at the usage point of each of those members without any need for creating new data structures or caching, I don't anticipate any real performance impact from the simple approach there. If it's discovered that is not the case during implementation then we can revisit that assumption.

Other than that, I thin the PR can be accepted as-is with the caveat that I may tweak it a bit in the future.

es-yuvanw:
@bfiete I've removed the temporary member resolution code, so project/workspace members values no longer change during compilation. Instead, custom build properties are now resolved in the existing `ResolveConfigString()` function, so any areas that already support resolution of built-in properties now support custom build properties as well. I've also added `ResolveConfigString()` calls for a few more members that we need for our build chain:

- Project paths
- Windows file version
- Windows product version

Ideally I would have liked to add support for property resolution for all project/workspace members, but since that will require a large number of additional property resolution calls, I decided to limited these changes to only what we need for our build chain, in order to reduce the blast radius of this PR. As always, I'm happy to make further adjustments if needed!

PR #2211, opened 2025-04-04
Implement UTF-8 aware string reversal

dadavadd:
Adds a Reverse(String strBuffer) method to String.
This method reverses the string, treating multi-byte UTF-8 characters correctly, and outputs the result to the given strBuffer.


bfiete:
I believe this can actually be done in-place and with a simpler solution:

Step 1: Iterate through the string and in-place reverse the individual 2, 3, and 4-byte utf8 encodings.
Step 2: Reverse the entire string

PR #2212, opened 2025-04-04
Implement String.Reverse() Method

dadavadd:
Adds an in-place Reverse() method to the String class, correctly handling multi-byte UTF-8 characters.


dadavadd:
Rewrote the method to be simpler, should be all good now :)

PR #2215, opened 2025-04-09
Added the Parallel class and the ForEach method in it

larkliy:


m910q:
I tested it, but had some issues. Probably related to the `Task` implementation.
It works fine in debug, but not release builds. The following seems to have an infinite loop going:
```beef
        int[2] items = .(0, 1);
        Parallel.ForEach<int>(items, scope (item) => { });
```

Some other thoughts on the code:
- If this goes into the corlib, it would be nice having it be available on non-Windows as well (Maybe use `Platform.BfpSystem_GetNumLogicalCPUs()`?)
- `tasks` can use `scope` instead of `new`. `task` can use `scope::`

bfiete:
I think it would be nice for someone to make a parallels library, but I'm not really thinking this is something that needs to go into corlib at the moment. 

System.Threading.Tasks is uncooked and probably should be taken of corlib as it stands, even...

disarray2077:
>System.Threading.Tasks is uncooked and probably should be taken out of corlib as it stands, even...

Yeah, I have also written a Parallel class before and IIRC the `Task' class has some thread race issues. For my project I ended up just writing my own simple Task class using `WaitEvent`.

Issue #557, opened 2020-10-05
[IDE][Enhancement] Expand and shrink selection


NicEastvillage:
A neat QOL feature for the IDE would be a keyboard shortcut to expand and shrink the current selection. For instance, from
![billede](https://user-images.githubusercontent.com/21122471/95129867-ef1cea00-075b-11eb-8ba9-f578b861c01b.png)
pressing the shortcut expands selection to
![billede](https://user-images.githubusercontent.com/21122471/95129908-02c85080-075c-11eb-85b1-04414449ad1d.png)
Pressing again expands to
![billede](https://user-images.githubusercontent.com/21122471/95129959-1f648880-075c-11eb-8cf3-46ff4a9d038d.png)
and then
![billede](https://user-images.githubusercontent.com/21122471/95129984-2ab7b400-075c-11eb-903b-f9f0d71fd901.png)
and so on. Shrinking goes in the opposite direction.

Essentially, the selection is walking up and down the abstract syntax tree based on where the cursor is. Jetbrains IDEs has a similar feature where the shortcut is `ctrl+W` for expand and `ctrl+shift+W` for shrink. However, `ctrl+W` is currently close tab, so I guess it has to be something else here.

NicEastvillage:
And for completion: From a selection like
![billede](https://user-images.githubusercontent.com/21122471/95130983-a9612100-075d-11eb-8fa4-4634466d9514.png)
expanding goes to the smallest expression that contains the previous selection. So
![billede](https://user-images.githubusercontent.com/21122471/95131195-f218da00-075d-11eb-9e21-b42a55d53d6c.png)

And shrinking always shrinks to the biggest selection smaller than the previous. So from
![billede](https://user-images.githubusercontent.com/21122471/95131330-27bdc300-075e-11eb-92c8-bdfb798db34e.png)
to
![billede](https://user-images.githubusercontent.com/21122471/95131363-3310ee80-075e-11eb-80a0-2594e1018ea2.png)

PR #2229, opened 2025-05-04
Feature: Multi-Cursors

kallisto56:
Hi there,

This pull-request brings a feature that allows for multiple cursors to be used in the same document. Hold `Alt` and press `LeftMouseButton` in a place where you want to put a cursor. Pressing `Escape` or just clicking somewhere (without holding `Alt`) will remove any secondary cursor.

This pull-request may be rejected, while parts or whole source code could be re-used to create a version of this feature that you think will be better for the project. Also, there is no rush or time-limit on accepting/rejecting this pull-request. I wish I could say that I'll assist in fixing bugs and consulting about related stuff, but lately I started experiencing internet problems (government firewall).

Given the amount of changes, I thought that it would be easier for you to understand it by providing an excerpt of a code:
```C#
class EditWidgetContent
{
	class TextCursor
	{
		// Primary cursor ALWAYS has an ID of zero
		int32 mId;

		EditSelection? mSelection;
		EditWidgetContent.LineAndColumn? mVirtualCursorPos;
		int32 mCursorTextPos;

		bool mCursorImplicitlyMoved;
		bool mJustInsertedCharPair;
	}

	// Special UndoBatch that can merge
	MultiCursorUndoBatchStart mMultiCursorUndoBatch;

	// Primary Cursor always sits at the front of the list
	List<TextCursor> mTextCursors = new List<TextCursor>() ~ DeleteContainerAndItems!(_);
	TextCursor mCurrentTextCursor;

	// Any widget that uses those fields, can keep doing that thanks to `ref`
	ref bool mCursorImplicitlyMoved => ref mCurrentTextCursor.mCursorImplicitlyMoved;
	ref bool mJustInsertedCharPair => ref mCurrentTextCursor.mJustInsertedCharPair;
	ref EditSelection? mSelection => ref mCurrentTextCursor.mSelection;
	ref int32 mCursorTextPos => ref mCurrentTextCursor.mCursorTextPos;
	ref LineAndColumn? mVirtualCursorPos => ref mCurrentTextCursor.mVirtualCursorPos;

	virtual void AdjustCursorsAfterExternalEdit(int index, int ofs, int lineOfs)
	{
		// Here we iterate over cursors and adjust their position with exception of the
		// currently selected one (unless this widget is not in focus).
		// 'lineOfs' has been added for cases, when cursor is in a virtual position
		// When modifying this method, test for consecutive selections, meaning:
		// 4 cursors that have selection and go one right after another.
	}

	override void MouseDown(float x, float y, int32 btn, int32 btnCount)
	{
		SetPrimaryTextCursor();
		if ((mIsMultiline) && (btn == 0) && (mWidgetWindow.GetKeyFlags(true) == .Alt))
		{
			mTextCursors.Add(new TextCursor(-1, mCurrentTextCursor));
		}
		else
		{
			RemoveSecondaryTextCursors();
		}

		// ...
	}


}

class EditWidget
{
	// Overriding EditWidget's KeyDown/KeyChar allows us to catch those
	// event before SourceEditWidgetContent and others
	override void KeyDown(KeyDownEvent keyEvent)
	{
		var ewc = Content;
		var isSingleInvoke = false;
		var keyFlags = mWidgetWindow.GetKeyFlags(true);

		if (((keyEvent.mKeyCode == (.)'Z') || (keyEvent.mKeyCode == (.)'Y')) && (keyFlags.HasFlag(.Ctrl)))
		{
			// Undo/Redo
			// This might be not ideal, since we are deleting secondary cursors only
			// for them to be created again during Undo/Redo, maybe some kind
			// of pool that keeps them for a while instead of deleting immediately
			ewc.RemoveSecondaryTextCursors();
			isSingleInvoke = true;
		}
		else if (keyEvent.mKeyCode == .Escape)
		{
			// For now, when we close AutoComplete by pressing Escape,
			// it unfortunately also remove secondary cursors.
			ewc.RemoveSecondaryTextCursors();
			isSingleInvoke = true;
		}

		ewc.RemoveIntersectingTextCursors();

		for (var cursor in ewc.mTextCursors)
		{
			ewc.SetTextCursor(cursor);
			base.KeyDown(keyEvent);
			if (isSingleInvoke)
				break;
		}
		
		ewc.SetPrimaryTextCursor();
		ewc.CloseMultiCursorUndoBatch();
	}

	override void KeyChar(KeyCharEvent keyEvent)
	{
		var ewc = Content;
		ewc.RemoveIntersectingTextCursors();
		for (var cursor in ewc.mTextCursors)
		{
			ewc.SetTextCursor(cursor);
			base.KeyChar(keyEvent);
		}

		ewc.SetPrimaryTextCursor();
		ewc.CloseMultiCursorUndoBatch();
	}
}
```

Additional notes:
- `MultiCursorUndoBatch` is created either somewhere down the call stack during `KeyDown/KeyChar`, or created and closed inside a single method, for example `SEWC.DuplcateLine()`.
- Pressing `Alt+Enter` when `QuickFind` is opened, selects all matches in the document or in a selection. Pressing `Shift+Alt+Enter` does the same, but the search is case-sensitive;
- Additional features: `Select Next Match` and `Skip Current Match and Select Next`;
- `Move Statement/Line Up/Down` removes secondary cursors for now;

Things, I'm not entirely sure I've done correctly:
- Changes in `AutoComplete.GetAsyncPos()` resolved issues with flickering of the AutoComplete's window;
- `EWC.CopyText()` and `EWC.PasteText()` inside IDE works fine, but I've just noticed that copying in IDE and then pasting to notepad for example has additional `\n` symbol at the end;

It was a pleasure working on that!

### DIFF
--- a/BeefLibs/Beefy2D/src/theme/dark/DarkEditWidget.bf
+++ b/BeefLibs/Beefy2D/src/theme/dark/DarkEditWidget.bf
@@ -243,9 +243,9 @@ namespace Beefy.theme.dark
 			return mLineCoords[anchorLine + 1] == mLineCoords[checkLine + 1];
 		}
 
-		protected override void AdjustCursorsAfterExternalEdit(int index, int ofs)
+		protected override void AdjustCursorsAfterExternalEdit(int index, int ofs, int lineOfs)
  		{
-			 base.AdjustCursorsAfterExternalEdit(index, ofs);
+			 base.AdjustCursorsAfterExternalEdit(index, ofs, lineOfs);
 			 mWantsCheckScrollPosition = true;
 		}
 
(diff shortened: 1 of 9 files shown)


MineBill:
Amazing work! I've been using it for a couple of hours and i love it! The only recommendation i have to automatically select the word under the cursor when `Select Next Match` is triggered and there is no active selection. Here is my half-assed attempt from my personal fork: https://github.com/MineBill/Beef/commit/cd22ae9fae70dc8b13b0daaeaee9a9379871ec63

kallisto56:
@MineBill, thanks! There is a feature request (https://github.com/beefytech/Beef/issues/557) that I believe will cover this case, although I'm not entirely sure how much time it will take me to make it.

bfiete:
I was comparing to Visual Studio Code and I noticed a couple things-

1) Is there any real value in showing the primary cursor in a different color from the secondary cursors? In VSCode they all just appear the same.. the blue cursors are feeling a little weird to me.
2) Did you add a way to add multiple cursors with the keyboard? I noticed in VS that Ctrl+Alt+Up/Down adds cursors, which seems pretty convenient.

bfiete:
I'm renaming these commands to match the VSCode names, which are "Add Selection to Next Find Match" (now defaulting to Ctrl+D), and "Move Last Selection to Next Find Match" which is supposed to be `Ctrl+K, Ctrl+D` but that's already taken by Reformat Document.

I can't quite understand the use case of "Move Last Selection to Next Find Match", and it does appear to behave somewhat differently in Sublime. How do you typically use that functionality?

bfiete:
Also, this functionality does interact with Quick Find in those editors but it does not properly interact with Quick Find in the Beef IDE so far...

kallisto56:
"Move Last Selection to Next Find Match" in VSCode moves primary selection from one match to the other and also removes any secondary cursors. For example, select first 'case' and then press `Ctrl+K, Ctrl+D`, which will lead to selecting second 'case'.
```C#
enum MyEnum
{
    case Foo;
    case Bar;
}
```
For the QuickFind, maybe I'm misunderstanding, but for the same example code above, you can press `Ctrl+F` and type 'case'. Then, pressing `Alt+Enter` will select all matches ignoring case. For case-sensitive search, you can press `Alt+Shift+Enter` instead.
Reason why QuickFind does not uses methods like `EditWidget.SelectNextMatch()` is because QuickFind may have a condition to search within selection.

kallisto56:
Correction: VSCode does not removes secondary cursors on "Move Last Selection to Next Find Match", it was just me pressing wrong hot-key.

bfiete:
This is on purpose, so you can actually *see* the cursors.

The actual bug is that you have Ctrl+D bound to two different things and the IDE is executing both of those actions every time you hit Ctrl+D

kallisto56:
With changes you've made, "Add Selection to Next Find Match" works almost exactly the same as in both, Sublime Text and VSCode, with exception that when new cursor is added, window is scrolled to that position. Basically, calling `EditWidget.EnsureCursorVisible()` for the newly secondary cursor. However, at the moment, it will not work, because `EnsureCursorVisible()` does that only for the primary cursor.

kallisto56:
One thing that I've noticed is that, both, Sublime Text and VSCode, do not scroll window to any cursor when user is typing text.

PR #2252, opened 2025-05-26
Expanded documentation support while hovering over a function

ESFraser:
## Context
Our team is currently working on improving the internal documentation throughout our codebase, And we've opted to use doxygen-style comments to add details and explanations to functions. While working on the documentation I've noticed that beef has very minimal support for said doxygen-style comments: just showning the brief description when hovered, or in the auto-complete list and just showing the parameters when typing them out.

## Solution
To that end, I thought it'd be generally useful to expand upon the behavior shown when a function is hovered and show various bits of detail about the function, instead of just the "brief", as shown below:
image

I have left the behavior for the description in the auto-complete list and the behavior when typing out function parameters untouched, as I believe they serve their purpose well as-is.

## Notes
If there are too many lines in the popup window, the vertical scroll bar widget will appear as expected but will _not_ allow any scrolling to occur: I believe this is due to the text internally being considered a `ListItem` as part of a `ListView`, and from what I can tell, the height of a list item is expected to be the height of one line in the displayed font, rather than something akin to `fontHeight * lineCount`. I am open to making any fixes required, but opted to first bring it up as a point of discussion as I am unfamiliar with the codebase as a whole and would like to to ensure the changes in this PR are easy to understand the reasoning behind rather than making what could be major changes and having to explain them in post
image

### DIFF
--- a/IDE/src/ui/AutoComplete.bf
+++ b/IDE/src/ui/AutoComplete.bf
@@ -17,6 +17,13 @@ namespace IDE.ui
 	{
 		public String mDocString = new String(256) ~ delete _;
 		public String mBriefString ~ delete _;
+		public String mAuthorString ~ delete _;
+		public String mReturnString ~ delete _;
+		public String mRemarksString ~ delete _;
+		public String mNoteString ~ delete _;
+		public String mTODOString ~ delete _;
+		public String mSeeAlsoString ~ delete _;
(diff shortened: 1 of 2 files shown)


hunterbridges:
@ESFraser now that this has been merged in, we should create an issue to track the tooltip scrolling bug you found

PR #2267, opened 2025-06-07
Add Documentation Reflection for Comptime

Rune-Magic:
```beef
/** @brief breif
 *
 *  text
 *
 *  @see HelloWorld.Program.GetDocString
 */
struct Foo
{
	/// defaults to 5
	public int bar = 5;

	/// does something
	public void DoSomething()
	{
		// ...
	}
}

[Comptime(ConstEval=true)]
static String GetDocString()
{
	Type type = typeof(Foo);
	String output = scope .();
	type.GetDocumentation(output);
	output.Append("\n", type.GetField("bar")->Documentation);
	output.Append("\n", type.GetMethod("DoSomething")->Documentation);
	return output;
}

Console.WriteLine(GetDocString());
```

prints:
```
/** @brief breif        *      *  text        *      *  @see HelloWorld.Program.GetDocString        */
/// defaults to 5
/// does something
```


bfiete:
Looks good, I think!

Issue #2264, opened 2025-06-06
Multi cursor: Issues with adding usings


m910q:
Hi

There are some issues when the IDE adds usings.
The cursor jumps to another place, and sometimes the using is added on the line above.

@kallisto56 confirmed it is an issue from multi cursor.

Clip of both cases:
https://github.com/user-attachments/assets/c090cdc3-023b-4850-b6b8-7e6922302048

kallisto56:
I think I got it. Will be in one of the upcoming pull-requests.

PR #2270, opened 2025-06-09
Fixed `FixIt` pasting in wrong place

kallisto56:
Hi there,

This pull-request fixes issue with `FixIt` pasting in wrong place, specifically `using` directive, as mentioned in issue https://github.com/beefytech/Beef/issues/2264.
There were two reasons why it was happening:
- Cursor position was assigned to `TextCursor` which did not updated virtual position;
- `AdjustCursorsAfterExternalEdit(...)` only updates primary cursor, when widget is not in focus, otherwise it assumes that cursor is in the right spot. This lead to cursor being in the wrong position after `FixIt` is applied;


m910q:
I can confirm the issue in #2264 is fixed

PR #2288, opened 2025-07-10
Surround System.gRand static initializer with #if !BF_RUNTIME_DISABLE

hunterbridges:
We were getting intermittent linker errors due to this static initializer in System.Random when building with runtime disabled


bfiete:
Those link errors should be fixed instead, but I'll punt by accepting this for now.

PR #2300, opened 2025-07-25
Free addrinfo after use

zerkawei:


bfiete:
It seems this is invalid, in the code block below:

```C#
sockAddr = addrInfo.ai_addr;
addrFamily = addrInfo.ai_family;
let addrlen = addrInfo.ai_addrlen;

addrInfo.ai_addr = null;
freeaddrinfo(addrInfo);
```

You are retaining `addrInfo.ai_addr` in `sockAddr` even after you free it through `freeaddrinfo`. I believe the correct solution is to change the method to return a struct that can be used and then be Disposed of.

bfiete:
I made a couple more changes @ https://github.com/beefytech/Beef/commit/163ddbf444922e9797bebad688ae571fe3ac84ed

1) This method is now called `ConnectEx`
2) There were still some cases where you were leaking the addrInfo from `getaddrinfo` during subsequent failures in Connect (ie: when `connect` itself failed)
3) This `Connect` never even used the incoming `port` value. Not sure how you were ever able to successfully connect to anything via this method

PR #2322, opened 2025-09-25
Allow linking emscripten's "fetch" API when building using wasm32

Starpelly:
This is just a draft for now, because I'm not sure I did this correctly.

image

It's self-explanatory; when using emcc to build, we toggle the "FETCH" flag if enabled. This is useful when you want to make HTTP requests in WASM or whatever. It was necessary to get [the leaderboards on this game working.](https://www.newgrounds.com/portal/view/998531)

This is probably a hack, and it doesn't even work unless you toggle it for every project in your workspace. Nevertheless, it worked for me, so this is definitely a useful feature to have.

### DIFF
--- a/IDE/src/BuildContext.bf
+++ b/IDE/src/BuildContext.bf
@@ -780,6 +780,9 @@ namespace IDE
 					if (!workspaceOptions.mRuntimeChecks)
 						linkLine.Append(" -s ASSERTIONS=0");
 
+					if (project.mWasmOptions.mEnableFetch)
+						linkLine.Append(" -s FETCH=1");
+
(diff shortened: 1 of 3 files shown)


bfiete:
To clarify further- the reason that `Enable Threads` is an option is that it has an effect on more than just the emcc arguments-  a different BeefRT library needs to be selected based on whether or not threading is enabled.

Starpelly:
Hey, good morning. I didn't understand what 'Other Linker Flags' meant; I've never used it before. I was working on 5 hours of sleep last night, just trying to get this working.

Honestly, that was my bad; this change was unnecessary. Sorry!

46bbw2zk26g51

PR #2325, opened 2025-09-29
Make ObseleteAttribute work on more targets

Rune-Magic:
it currently doesn't work on fields


bfiete:
Was there a reason you didn't just add the flags that needed to be added? IE: `Field` and `StaticField`?

Rune-Magic:
it's more generic, i.e. function, delegates pointers also weren't supported, etc.

PR #2369, opened 2025-12-18
Fix compiler slowdowns due to errant entries in mFunctionsUsingSimd

hunterbridges:
A couple of months ago, we started hitting compiler hangs on Release mode builds of our Editor project when some of our team members were trying to update to the latest nightly. I had the team revert to the last known working version, which for us was the `2025-07-13` nightly.

Now I had a chance to dig into the issue. I managed to figure out that the issue was introduced between the `2025-07-28` nightly and the `2025-07-29` nightly. 

I am fairly certain the issue was introduced in this SIMD-related changeset: https://github.com/beefytech/Beef/commit/a33d046bfe2c66a4e2ec28281c2f38e34ce60566

It looked like some worker threads were spending quite a lot of time in `ApplySimdFeatures()`. `mFunctionsUsingSimd` ended up having a ton of entries, even when I turned SIMD off for the workspace. It turned out that "None" or "Not Set" SIMD types were being considered as "functions using SIMD", which had downstream implications in the compilation process.

This change basically omits functions from `mFunctionsUsingSimd` when the SIMD type is None or Not Set.

I still think something is amiss with the SIMD stuff added in the above changeset, because even with the fix here I have to set `Workspace > Beef > SIMD Instructions` to `None` in Release mode for it to go back to the faster compile times in `2025-07-28` Nightly and earlier.

### DIFF
--- a/IDEHelper/Compiler/BfIRCodeGen.cpp
+++ b/IDEHelper/Compiler/BfIRCodeGen.cpp
@@ -3449,13 +3449,12 @@ void BfIRCodeGen::HandleNextCmd()
 			result.mValue = func;
 			SetResult(curId, result);
 
+			func->addFnAttr("no-trapping-math", "true");
+			func->addFnAttr("min-legal-vector-width", "0");
+			func->addFnAttr("tune-cpu", "generic");
+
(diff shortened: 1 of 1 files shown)


hunterbridges:
alrighty, moved those attributes back out

PR #2377, opened 2025-12-24
Fix iOS and macOS x86 linking

Fusioon:
- FIxes BeefRT building for iOS and macOS(x86_64)

- Adds `BF_MACHINE_X64` and `BF_MACHINE_AARCH64` when buidling with BeefBoot, then the arch is stored in `IDEApp.sArch64Name` and used for target triple selection when targeting macOS

- add `--disable-docs` to _libffi_ `./configure` arguments, this is to fix build errors when machine doesn't have installed `texinfo` and `automake` - this issue seems to be caused by files in `libffi/docs` having different timestamp than expected when pulled from github


Fusioon:
done

PR #2396, opened 2026-02-15
Add Poll to Sockets

zerkawei:
This PR adds poll() support to Sockets given it's recommended over select() for modern applications.

### DIFF
--- a/BeefLibs/corlib/src/Net/Socket.bf
+++ b/BeefLibs/corlib/src/Net/Socket.bf
@@ -608,7 +608,6 @@ namespace System.Net
 		public ~this()
 		{
 			if (mHandle != INVALID_SOCKET)
-				shutdown(mHandle, .Both);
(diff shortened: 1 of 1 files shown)


bfiete:
It looks like this won't work on Windows, as `WSAPoll` is not called by `Poll`, only the Linux `poll` is...

bfiete:
Given that Socket is a "thin" API wrapper, I think it's probably better to split out the `Close` and `Shutdown` into separate methods. 

I believe the destructor should only `Close`, and if `Shutdown` behavior is desired then the user should do that himself.

zerkawei:
How should `Shutdown` interact with `IsConnected` ? 
Do we keep track of which side of the connection has been shut down ? 
Do we only update `IsConnected` with `Close` ?
Do we only permit shut down of both sides at once ?

Issue #2403, opened 2026-03-15
[Enhancement] Build Classes


Rune-Magic:
Okay, so one of beef's main selling points is developer productivity and for the most part that's true as long as you don't leave the .bf files. But currently in order to add prebuild and postbuild logic you have to leave the .bf files.
Typing into the tiny UI elements and managing configs and platforms in just a pain. So the idea is to move this logic into a .bf file and run it using the comptime machine similar to the IDE file generator. I am thinking something like this:
```cs
class BuildLogic : Compiler.BuildLogic
{
    public override void PreBuild() { ... }
    public override void PostBuild() { ... }
}
```

PR #2408, opened 2026-03-21
Build Logic Classes

Rune-Magic:
This adds build logic classes as described in #2403, progress:
- [x] prebuild and post build code runs
- [x] console output
- [x] debugging support
~~dynamically add additional lib paths from build logic~~


Rune-Magic:
https://github.com/beefytech/Beef/pull/2417

PR #2417, opened 2026-03-25
Comptime Improvements

Rune-Magic:
adds:
- ~~`Console.Write[Line]` to comptime~~
- ~~`Environment.[G|S]etVariable` to comptime~~
- make `Compiler.Options.*` eval to their enum type instead of to `int32`
- add `Console.RunCommand`


Rune-Magic:
```
ERROR: Unqualified dot syntax cannot be used with type 'int32'
   case .Og, .OgPlus:
         ^^
```

PR #2429, opened 2026-04-09
Prevent IDE crash for malformed recovery DB

zerkawei:
This PR fixes a crash that happened because of an unhandled error in the recovery db read in case of malformed data.


bfiete:
Also note that the dataSize reading line could have been written as `if (int32 dataSize = fs.Read<int32>())` which I personally find nicer.

PR #2430, opened 2026-04-19
Allow 'none' compilation target

hermansimensen:
Support 'none' target for bare-metal compilation (e.g. thumbv8m.main-none-eabi)


bfiete:
Seems reasonable.

PR #2440, opened 2026-05-19
Hook up LLDB debugger backend on linux

mdsitton:
With this the new lldb backend atleast works in linux now.

### DIFF
--- a/IDEHelper/CMakeLists.txt
+++ b/IDEHelper/CMakeLists.txt
@@ -123,8 +123,6 @@ file(GLOB SRC_FILES
     Targets.cpp
     X86XmmInfo.cpp
 
-    LinuxDebugger.cpp
-
     Beef/BfCommon.cpp
     Clang/CDepChecker.cpp
     Clang/ClangHelper.cpp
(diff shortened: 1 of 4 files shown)


bfiete:
This appears to break macOS:
```
In file included from /Users/runner/work/Beef/Beef/IDEHelper/LinuxDebugger.cpp:2:
/Users/runner/work/Beef/Beef/IDEHelper/LinuxDebugger.h:3:1: error: unknown type name 'NS_BF_DBG_BEGIN'
    3 | NS_BF_DBG_BEGIN
      | ^
/Users/runner/work/Beef/Beef/IDEHelper/LinuxDebugger.h:8:14: error: expected ';' after top level declarator
    8 | NS_BF_DBG_END
      |              ^
      |              ;
/Users/runner/work/Beef/Beef/IDEHelper/LinuxDebugger.cpp:12:1: error: a type specifier is required for all declarations
   12 | USING_NS_BF_DBG;
      | ^
3 errors generated.
```

bfiete:
Also I don't think there's a specific reason you needed to introduce a LinuxDebugger.cpp/h, I think it probably would have been best just to put it inside LLDBDebugger but put a conditional '#idef !defined BF_PLATFORM_WINDOWS` around it since that's the only exception to when you would want that code if LLDB_ENABLED were defined.

bfiete:
Oh right there already *was* a LinuxDebugger. I would suggest deleting that, actually.

PR #2448, opened 2026-05-23
Initial remote debugging support (LLDB)

hermansimensen:
Adds the option to debug beef software on embedded platforms through the gdb-remote protocol.

It adds a button under the debug menu, "Connect to remote target". This is only available with the LLDB debugger, otherwise it will be greyed out.

Some things to note are:
- I swapped out the liblldb.dll with the one from llvm's 22.1.5 github release because it is compiled with libxml2 support. This is important for reading target definition files from remote servers. Remote debugging can still work without this, depending on the platform, but will use a fallback solution that might result in incomplete register and thread information. I have not updated the build script to build liblldb with libxml2 because I don't know your stance on it.
- Currently you have to supply an elf file to load/debug. The beef IDE cannot generate this file on its own, but I've created mine through a linker script specifically for the target I'm working from. I have not planned to implement this functionality. 

I've tested the changes with an OpenOCD server, specifically this[ patch. ](https://github.com/raspberrypi/openocd/commit/acff23ffd100479d6481a1155123774b79a990b9)

remotedebug

### DIFF
--- a/IDE/src/Commands.bf
+++ b/IDE/src/Commands.bf
@@ -361,6 +361,7 @@ namespace IDE
 			Add("Zoom Out", new => gApp.Cmd_ZoomOut);
 			Add("Zoom Reset", new => gApp.Cmd_ZoomReset);
 			Add("Attach to Process", new => gApp.[Friend]DoAttach);
+			Add("Connect Remote Debugger", new => gApp.[Friend]DoConnectRemote);
 			Add("Move Last Selection to Next Find Match", new => gApp.Cmd_MoveLastSelectionToNextFindMatch);
 
 			Add("Test Enable Console", new => gApp.Cmd_TestEnableConsole);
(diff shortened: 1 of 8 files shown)


bfiete:
Some things have changed since this initial PR.

Now GDB support is added, and there are more ways to launch that to remote targets. I ended up encoding the remote target specifiers in the launchPath:

path@gdb_wsl - Launches GDB through WSL
path@gdb_ssh:server - Launch GDB through SSH connection
path@gdb:host:port - Connect to GDB server on host:port. Path is ignored and can be blank.

Those launchPath encodings also allow for the correct debugger to be selected by the IDE.

Also when buildng IDE UI's, make sure they scale properly with UI scale. You can test most easily with Ctrl+Plus, Ctrl+Minus, Ctrl+0. Generally you can just wrap `GS!(x)` around those hard-coded constants and it just works.

hermansimensen:
Closed in favor of reworked #2468.

PR #2460, opened 2026-06-12
fix Comptime Output only displaying once compiling finishes

### LINKED ISSUE #2454: comptime logs are hidden until compiling finishes
`Console.Write[Line]` works in comptime, but the results are only displayed once compiling finishes

Rune-Magic:
fixes #2454


bfiete:
Can you separate this into different PR's: the typedef changes vs the ComptimeOutput changes

Rune-Magic:
> Can you separate this into different PR's: the typedef changes vs the ComptimeOutput changes

done

bfiete:
This has been resolved in a more general way at 1e79f328169d46d53115a0e5abe0b641b10a69f7

It has the added benefit of showing compile errors as they occur, and leaves the compiler output stream in order (ie: console output appears at the appropriate location relative to other errors and warnings)

PR #2464, opened 2026-06-14
Filter options for errors panel

aharabada:
This has been on my wishlist for ages, and I finally found some time to make it happen.

This PR adds several filter options to the errors panel that you usually find in other IDEs.

The following filter options were added:
1. A drop-down to select the scope of errors:
	- __Entire Workspace__: Shows all errors
	- __Current File__: Only shows errors for the document that is currently focused
	- __Current Project__: Only shows errors for the project that the "current file" belongs to
	- __Open Files__: Only shows errors for all documents that are currently opened
2. Toggle buttons to show or hide errors or warnings
   These buttons also show how many errors and warnings the entire workspace has, as well as how many match the current filter settings.
3. Search box to filter error messages by text. This will match:
	- Description
	- Project name
	- File name
	- The number in the error-code column
4. Allow sorting the error list by clicking the column-headers
	- Shift-clicking allows sorting by multiple columns at the same time
	- implemented default order: Project -> File -> Line (which is the default for Visual Studio)

# Images:
Here are some images:
ErrorsPanel

__Only show errors:__
ErrorsPanel_onlyErrors

__Text search:__
ErrorsPanel_textSearch

__Error code search:__
ErrorsPanel_searchCode

# Open Questions:
1. I have a question about the `SearchEditWidget`: It has a hard-coded content shift to make space for the search icon. However the icon is rendered "manually" in `PropertiesDialog.DrawAll`:
https://github.com/beefytech/Beef/blob/8aeb8602bb4f2875401fa9b8e63fb0b364c068b0/IDE/src/ui/PropertiesDialog.bf#L2116-L2119 I mimicked this behavior in ErrorsPanel.DrawAll. I was wondering, if there was a reason to not override the `Draw`-Method for `SearchEditWidget` and render the icon in there?
2. I merged parse and resolve errors into a single sorted list before generating the list view items. Previously, parse errors always appeared above the resolve errors. Was that original top-ordering intentional? If so, I can restore it for the default sort order. https://github.com/beefytech/Beef/blob/8aeb8602bb4f2875401fa9b8e63fb0b364c068b0/IDE/src/ui/ErrorsPanel.bf#L361-L376

### DIFF
--- a/BeefLibs/Beefy2D/src/theme/dark/DarkButton.bf
+++ b/BeefLibs/Beefy2D/src/theme/dark/DarkButton.bf
@@ -11,6 +11,8 @@ namespace Beefy.theme.dark
     {
         public String mLabel ~ delete _;
 		public float mDrawDownPct;
+		public FontAlign mLabelAlign = .Centered;
+		public float mLabelXOfs;
 		public float mLabelYOfs;
 
         [DesignEditable(DefaultEditString=true)]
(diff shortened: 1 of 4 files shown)


bfiete:
1) Hm, I'm not quite sure why I did it that way. Either there was a good reason that I can't recall or I made the wrong choice there.
2) The idea would have been that resolve errors can result from earlier parsing errors, so I considered parsing errors to be possibly a kind of "root" error and thus more important, and was prioritizing those by showing them first. I'm not particularly tied to that reasoning, however, and I don't even use the error panel personally.

bfiete:
Seems pretty good! I made a small tweak to the combo box size so it matches the others (ie: platform and config selector on the lower right)

aharabada:
Ah, good catch. They did match at some point, but I had changed the toolbar height kind of last minute

PR #2476, opened 2026-07-20
Static theme coloring

Grimelios:
Adds a "Static" theme option that colors static variable declarations and their usages. "Static" is a subset of "Member".

Adjusting settings to parse the new color was trivial. Coloring static declarations was also straightforward, utilizing `propertyDeclaration->mStaticSpecifier` in `BfSourceClassifier.cpp`. For coloring usages of static variables, I modified `BfExprEvaluator.LookupField` to optionally provide the relevant `BfFieldDef` via pointer.

I spot checked a number of files touching static variables in various ways, and everything appears to color correctly. Here's an example from `SourceEditWidgetContent.bf`, where members are green and statics are red.

image


bfiete:
I had to revert some code in BfExprEvaluator.cpp - it failed tests in test_build.bat

I was unable to trivially resolve the problem. I can't guarantee I'll look into it very hard, given this is a low-priority feature.

Grimelios:
Understandable. I ran into a crash late last night with my PR merged that I didn't encounter during testing, so I was already planning on investigating further.

PR #2486, opened 2026-08-17
Add an option to make autocomplete suggest unimported namespaces

ThatCreeper:
I have not fully tested this.

All this PR does is add a setting in the menu ("Autocomplete Ignores Namespaces", defaulting to off) that disables the namespace checks in the autocomplete engine. Anything completed using this will show up as an error, because it doesn't add the correct using statement.

I'm not sure what the best way to make it add a using statement would be. My first thought would be to add a step after the completion that checks for and adds a missing using statement, but it would also probably be possible to leech off of Fixit.

### DIFF
--- a/IDE/src/Compiler/BfParser.bf
+++ b/IDE/src/Compiler/BfParser.bf
@@ -105,6 +105,7 @@ namespace IDE.Compiler
 		public int32 mTextVersion = -1;
 		public bool mIsUserRequested;
 		public bool mDoFuzzyAutoComplete;
+		public bool mShouldAutoCompleteIgnoreNamespaces;
 		public Stopwatch mStopwatch ~ delete _;
 		public ProfileInstance mProfileInstance ~ _.Dispose();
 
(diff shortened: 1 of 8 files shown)


bfiete:
There *is* some existing precedent here: look at how VS handles it for C#, which is an individual toggleable setting on the autocomplete.

The key insight would be that this is generally not something you *always* want on, it's something you'd want on when you do not yet have the correct namespaces set up for a given source file.

bfiete:
Also, please *do* make sure that your PRs are properly tested and provide a complete solution, not a partially-implemented solution. If you want to code only part of a solution then you could always ask someone to make a PR to your own repo to help you finish it before you submit it to the main repo.

ThatCreeper:
I've used this PR in my own Beef project and in the IDE project, and it's seemed to work fine. I just don't know if there's any edge cases I haven't hit or peculiarities with my system, which is what I mean when I say I haven't fully tested it. I'm planning on building it on a second computer soon.

This feature is gated behind an option in my patch. The differences from Visual Studio 2026's implementation (as I can see), are that...

1. Visual Studio defaults the feature to on.
2. Visual Studio refers to it as "Show items from unimported namespaces" whereas I've called the setting "Autocomplete Ignores Namespaces" (Perhaps "Autocomplete Includes All Namespaces" would be better, but it might be a little long? EDIT: Perhaps "Autocomplete Unimported Items"?)
3. Visual Studio has an option "Completion behavior for items from unimported namespaces" which determines if a using statement is added when you complete. My implementation does not yet have that feature and does not have the setting.
4. Visual Studio distinguishes unimported symbols by putting the namespace in gray next to the autocomplete item.

C#, to my knowledge, doesn't have an equivalent of a file-scope static block, so I'm not totally sure what to do with that.

Also, I was debating whether or not to mark the PR as a draft because of the lack of using generation, but automatically adding a using statement seems like a much more complicated change, and just this alone already feels helpful to me.

ThatCreeper:
In order to make autocomplete generate using statements, and in order for it to show the namespace in gray like Visual Studio, the namespace information would have to be hoisted from C++ to Beef.

I am not familiar enough with the data structures and the serialization to feel comfortable building a good implementation of this. As is, though, the PR seems to work fine.

EDIT: I built a bad simulacra of the using statements feature in [this branch](https://github.com/ThatCreeper/Beef/tree/autocomplete-using-bad), which is not going in the PR. It is buggy and gets the namespace by parsing the autocompletion documentation field, but it might be interesting to someone.

PR #2490, opened 2026-08-24
Fix break in nested scope inside a switch case

### LINKED ISSUE #2479: Nested switch break can cause EXCEPTION_ILLEGAL_INSTRUCTION
Despite a nested `break;`, a switch is still considered to never break when they all return a value.
Causes `EXCEPTION_ILLEGAL_INSTRUCTION` in Debug and crashes in Release.
Tested in latest nightly.

```beef
class Program
{
	{
        switch (0) {
            case 0:
                if (true) { break; }
                return 1;
            default:
                return 2;
        }
		return 0; // <-- "Unreachable code" warning in IDE
	}
}
```

m910q:
Fix setting `allHadReturns = false;` when a `break;` appears in a nested scope.

I'm not very familiar with how the compiler works, so it's an attempt that I hope can save you some time.

AI used to identify and make a bad fix
Human used to improve fix and do some manuel testing.

Fixes #2479

This PR does not fix how `break;` seems to be treated differently than `return;` when it comes to unreachable code within a single switch case. Compile time known branches works differently for instance:
`if (compTimeFalseValue) { return; }` is ignored as a return, but a `break;` is seen as a possibility, as far as I can tell.

### DIFF
--- a/IDEHelper/Compiler/BfStmtEvaluator.cpp
+++ b/IDEHelper/Compiler/BfStmtEvaluator.cpp
@@ -5161,8 +5161,8 @@ void BfModule::Visit(BfSwitchStatement* switchStmt)
 			openedScope = false;
 			deferredLocalAssignDataVec[blockIdx].mHadReturn = hadReturn;
 			caseCount++;
-			if ((!hadReturn) &&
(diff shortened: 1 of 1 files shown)


bfiete:
Can you provide some test methods that fail with the current implementation but pass with this fix?

PR #2494, opened 2026-09-05
SDL: implement BFWindow_CaptureClientBits

jayrulez:
The export was added for the Windows backend only, so the SDL backend failed to link once UICapture.bf started calling it. Reads the framebuffer back with glReadPixels, reverses the rows to match the top-down order the Windows path produces, and forces alpha opaque because a context without an alpha channel reads back zero.

Note: This fix was implemented with Claude. 
Tested on Ubuntu 26.04.


bfiete:
The implementation has changed and this will now no-op on Linux. You can either adjust the PR to properly work or accept the no-op. It's only useful for the MCP.

jayrulez:
Rebased and adjusted.

PR #2495, opened 2026-09-05
Compiler: populate the aliased type when resolving through a typealias

jayrulez:
ResolveTypeResult walks a chain of aliases, populating each alias and then stepping to what it refers to, but does not apply the caller's requested populate level to the type it ends up returning. The unpopulated defaults are 0 for size and 1 for align, so the result is a silently wrong constant instead of an error.

```C#
    struct S { public float a; public float b; }
    typealias A = S;
    sizeof(A)   // 0, want 8
    alignof(A)  // 1, want 4
```

Affected, all silent: sizeof, alignof and strideof through an alias, through an alias chain, to a generic instantiation, to an alias declared inside a class, to an enum, to a tuple, to a type in another namespace, and the same read from a const initializer or a comptime block.

Not affected, because these reach the type through paths that populate it anyway: offsetof, a field or sized-array field typed by the alias, sizeof of an array of the alias, a generic parameterised by the alias, and typeof(alias).Size and .InstanceSize.

Add Aliases.TestAliasedTypeAttrs over a plain struct, an alias chain and a struct whose body is emitted at comptime.

Note: This was debugged by and the fix suggested by Claude.

### DIFF
--- a/IDEHelper/Compiler/BfModuleTypeUtils.cpp
+++ b/IDEHelper/Compiler/BfModuleTypeUtils.cpp
@@ -10320,6 +10320,7 @@ BfType* BfModule::ResolveTypeResult(BfTypeReference* typeRef, BfType* resolvedTy
 	if ((populateType != BfPopulateType_TypeDef) && (populateType != BfPopulateType_IdentityNoRemapAlias))
 	{
 		int aliasDepth = 0;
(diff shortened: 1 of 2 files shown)


bfiete:
Looks reasonable, and I appreciate a reasonable-looking test as well.

PR #2498, opened 2026-09-06
BeefySysLib: implement BfpFile_GetTempPath on POSIX

jayrulez:
It was NOT_IMPL, so Path.GetTempPath returned an empty string on Linux and macOS and every caller composed a path at the filesystem root. corlib's own FileTests.WriteAll_PropagatesResult is one of them, and could not have passed on either platform.

The temp directory was already known to this file: BfpFile_GetTempFileName sits directly below and hardcodes /tmp. This reads TMPDIR first, as POSIX tools do, falls back to the same /tmp, and returns it with a trailing separator to match what the Windows implementation returns from GetTempPathW.


bfiete:
You got pipped by this https://github.com/beefytech/Beef/pull/2492
Resubmit an adjusted PR if you think there's something to add still.

PR #2499, opened 2026-09-06
corlib: FileMode.Create must truncate in BufferedFileStream

jayrulez:
System.IO.FileStream derives from BufferedFileStream, and BufferedFileStream carries its own copy of the FileMode to BfpFileCreateKind mapping. That copy has drifted from UnbufferedFileStream's: Create sets CreateAlways without the Truncate flag, so opening an existing file with FileMode.Create keeps its contents and a shorter write leaves the tail of the previous one behind. FileMode.Create is documented as overwriting the file, and every caller of FileStream gets the buffered class, so the working spelling is the one nobody reaches.

Truncate had drifted the other way, to CreateAlways, which creates a file that is documented to fail when it does not exist. Both cases now read as UnbufferedFileStream has them.

The tests use working-directory relative names rather than Path.GetTempPath, because BfpFile_GetTempPath is NOT_IMPL on POSIX.

### DIFF
--- a/BeefLibs/corlib/src/IO/FileStream.bf
+++ b/BeefLibs/corlib/src/IO/FileStream.bf
@@ -405,12 +405,13 @@ namespace System.IO
 				createKind = .CreateIfNotExists;
 			case .Create:
 				createKind = .CreateAlways;
+				createFlags |= .Truncate;
 			case .Open:
 				createKind = .OpenExisting;
 			case .OpenOrCreate:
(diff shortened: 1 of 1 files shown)


bfiete:
Did you test on Linux only? The tests failed on Windows - which was actually a platform bug, not a bug with your implementation. Fixed now.

jayrulez:
Oh yes, sorry I only tested on Linux. I'll remember to check windows next time.

PR #2505, opened 2026-09-17
IDE: the emscripten compiler name is not emcc.exe off Windows

jayrulez:
BuildContext appended "upstream/emscripten/emcc.exe" unconditionally, so a wasm32 link on Linux or macOS failed with "Failed to execute ...emcc.exe" after the Beef compile had already succeeded. The emsdk ships emcc.exe only on Windows.


bfiete:
I don't think Linux or macOS ever worked with wasm, did it?

jayrulez:
It's my first time trying it there. With these few fixes it works.

PR #2507, opened 2026-09-17
IDE: a configured EmscriptenPath satisfies the wasm toolchain check

jayrulez:
The is a check for wasm/EmsdkDep2_Done.txt. That marker ships inside EmsdkDep2.zip and is unpacked by fetch_wasm.bat, a Windows batch file, so no POSIX checkout can produce it by any route. 

Check instead whether there is an emcc where the setting points. GetEmccPath builds that path for both the check and the command line.


bfiete:
No, we cannot get rid of the EmsdkDep2_Done.txt check on Windows

jayrulez:
Updated to check both the marker file and the emcc executable. Is this okay now?

bfiete:
Still wrong. EmsdkDep2_Done.txt is the mechanism that forces Emscripten to update on Windows. This code would make Emscripten never update.

jayrulez:
I will get back to this when I am back on Windows in a few days.

jayrulez:
Another PR handles this differently:
https://github.com/beefytech/Beef/pull/2523

jayrulez:
Closing as this was addressed in the merged PR referenced above.

PR #2509, opened 2026-09-17
BeefRT: a dynamic call for wasm, so reflection Invoke works there

jayrulez:
MethodInfo.Invoke is the only dynamic call Beef has, and it goes through FFILIB. The wasm runtime is built with BF_DISABLE_FFI, where PrepCif return FFIResult_NoFFI and Call is no-op, so every Invoke on wasm failed.

Symptom: Type.GetCustomAttribute has to construct the attribute to return it, so an attribute plainly present read back as absent, while HasCustomAttribute (a type id compare, no call) still said yes.

wasm does not need libffi. libffi is complicated because ABIs like SysV x86-64 scatter a struct's fields across registers, so a call has to be assembled per signature. clang's wasm32 ABI passes EVERY struct indirectly as one pointer and returns one through a hidden sret pointer prepended to the arguments: measured on the table entries, an 8 byte and a 24 byte struct both take arity 1, and returning either takes arity params+1. So every case is one number per argument, which is what a call through the function table already does.

PrepCif fills the cif record, Call maps each FFIType.TypeKind to a slot kind and dispatches through EM_JS to wasmTable.get(fp).apply. BuildContext exports wasmTable on every wasm link, without which the runtime's JS cannot see it.

BF_DISABLE_FFI still means no libffi rather than no FFI, so it keeps working for any other target without one. ClosureAlloc stays unimplemented, and its comment now says why: corlib binds the allocator and not ffi_prep_closure_loc.

Covers the shapes that can go wrong: i32, f32, f64, an i64 past 32 bits, a struct by value, a 24 byte struct by value, a struct returned by value, a void return, and the GetCustomAttribute case that surfaced it.


jayrulez:
I'll rebase this

bfiete:
Well, I don't really love there being a big block of emscripten-specific code thrown into a core BeefRT file. At minimum it needs to be separated off into some wasm-specific support file. I don't love that I don't have any idea whether this is actually a full and proper implementation, as it seems there could easily be cases it doesn't cover... but I guess that's still better than zero invocation support.

jayrulez:
Refactored

PR #2516, opened 2026-09-19
Expose parameter defaults to comptime reflection

jayrulez:
MethodInfo.GetParamDefault returns the default as written in the declaration, empty when there is none; HasParamDefault tests it. The comptime param info carries the initializer's source text, so a generator can paste it into emitted code. Runtime parameter data still has no defaults and answers empty.


bfiete:
This doesn't really feel right - it appears to expose functionality that is does not provide. 

A user would find this when looking for a way to retrieve the default value for a method invocation. This does not do that- it doesn't work at runtime *at all*, and it doesn't even work at compile time, since the text probably won't even compile because it would be evaluated in the wrong context (ie: different class, different project, different included namespaces, etc.). I'm not even sure why you'd want this at all, but it screams "dirty hack" to me.

It COULD be appropriate if we had a more fully realised "TypeDeclaration" support with MethodDeclarations and FieldDeclarations etc, but we don't.

bfiete:
What *is* your use case, anyway? Are you trying to "skip over" a default or something?

jayrulez:
Good points, it is indeed hackish.

I have a comptime binding generator ([OnCompile(.TypeInit), Comptime] in a "surface root" project) that walks every "[Scriptable] "method in the root's dependency closure and emits a thunk per method for a scripting VM (AngelScript at the moment).

I need default values for 2 things:


Arity. A method like Spawn(Guid prefab, Float3 position, Quaternion rotation = .Identity, EntityHandle parent = .Invalid) needs to be callable from script as Spawn(guid, pos), so the generator has to know how many parameters are required and register one script overload per arity. That only needs to know whether a parameter has a default.

The value when the script omits the argument. That's what I was using the text for: the thunk declares a local of the parameter's (fully qualified) type and pastes the initializer. It works for literals and dot-shorthands like .Identity.  this is the hacky part and not generally usable. Unqualified name from the declaring scope will fail here.

Thinking about it some more, the second one doesn't need the text. The thunk can emit one call per arity (switch (argCount) { case 2: Spawn(a0, a1); case 3: Spawn(a0, a1, a2); ... }) and let overload resolution evaluate the default in the declaring context. Then the only thing reflection needs to know is whether the parameter is optional. This could be made to work at comptime and runtime. This is similar to the existing Splat/Params flag. 

It could be generally useful outside my binding generator.


Would you be open to that instead: a HasDefault bit on the param flags, exposed as MethodInfo.HasParamDefault(int).

bfiete:
Yes, I would be happy to have an extra flag for that on ParamFlags. I would just leave it accessible through GetParamFlags though, not an extra MethodInfo method

PR #2517, opened 2026-09-20
Return a file time from BfpFile_GetTime_LastWrite on POSIX

jayrulez:
A BfpTimeStamp is a Windows file time: 100ns ticks since 1601, which is what the Windows implementation and BfpFindFileData_GetTime_LastWrite return and what File.GetLastWriteTime decodes. The POSIX path returned raw st_mtime seconds, so File.GetLastWriteTime(path) read as a moment in 1601 while the same file's enumeration entry read correctly. Convert through BfpToTimeStamp like the enumeration does, with the Darwin st_mtimespec spelling. A corlib test pins the two paths to each other.


bfiete:
We can get rid of this test too, it's clear enough. And we don't really want tests that write files to the filesystem if possible.

jayrulez:
test dropped

PR #2518, opened 2026-09-20
Populate the type an alias names when comptime reflects the alias

jayrulez:
CeContext::GetReflectType populated the alias and then built its type data, which reads through to the aliased type: WantsGCMarking on an alias forwards to it. In a resolve-only pass the aliased type can still be undefined at that point, which asserts in a debug build and carries on with an undefined type in a release one. Populate the target too.

Seen from an [OnCompile(.TypeInit)] generator reflecting a typealias to a function type in the IDE's background pass.


jayrulez:
Same question here @bfiete . Should I drop these tests from the PR?

bfiete:
Nah these are ok - they are more like proper "general" tests rather than laser-targeted tests for one specific bug.

bfiete:
Except now this has merge conflicts.

jayrulez:
Yeah, touched the same line as a previous. I will rebase

jayrulez:
Rebased

PR #2519, opened 2026-09-20
Mangle assignment operators by name on the GNU side

jayrulez:
The GNU mangler had no case for an operator declaration's assignment op, so every 'operator+=', 'operator-=' and the rest mangled to an empty method name and collided whenever their parameter types matched. The collision was papered over by the function index, but the debug info kept the shared linkage name, and a debug compiler's LLVM verifier rejects a module where two subprograms claim one function: "conflicting debug info for argument". Seen on a struct with inlined += and -= over the same type, where the release build had been fine all along.

The Itanium codes are used, as the binary and unary operators already are; '<=>' gets 'ss' for the same reason. The MS mangler already named these.


jayrulez:
@bfiete  Drop these tests too?

PR #2520, opened 2026-09-20
Leave an undef constant alone when folding '~'

jayrulez:
CreateNot folded every constant operand, unlike CreateNeg and the binary folds, which return an undef unchanged. For an undef the mTypeCode read is the union tag, BfConstType_Undef, and the payload inverted is the undef's type, so the fold produced a constant tagged undef whose type kind was the complement of the real one: TypeKind_None, id -30 for an int32. CastToValue then asserted on it in a debug compiler and carried on with garbage in a release one.

Reached from a generic method taking alignof(T) into a 'let' and masking with ~(align - 1), which is what a subobject writer does.


jayrulez:
Compiler hard crashes on my project without this fix.

bfiete:
Yeah, I don't think we need a test for this one.

jayrulez:
So drop the test?

bfiete:
Yeah, I don't want to flood the compiler with a bunch of tests that are checking things that have zero chance of ever breaking again.

jayrulez:
Done

PR #2521, opened 2026-09-20
Keep a module's interface slot dependency across an extension

robknopf:
Replaces the earlier version of this PR, which dropped the `>= 0` guard in `UpdateRevisedTypes`. That was wrong for the reason given in review: it rebuilt every module that had never recorded a slot dependency, most of which genuinely have none. This version leaves that condition exactly as it is and fixes where the dependency was actually being lost.

### What was happening

`mUsedSlotCount` is set by `HadSlotCountDependency()` when a module emits a virtual or interface call, and normalized to the current slot count when the module finishes. `-1` means the module never recorded a dependency, and the rebuild check rightly skips it.

`StartExtension()` re-enters a finished module to append on-demand methods. It calls `Init(false)`, and `Init` resets `mUsedSlotCount = -1` unconditionally. But an extension keeps the code already generated — so the dependency that code recorded was discarded while the code stayed.

Tracing the flag on the module that crashes, `List<ErrorHandlerData.ErrorHandler>`, across one compile:

```
INIT    rev=1  -1 → -1   fullRebuild=1
MARK    rev=1  -1 →  0   in Add          virtual dispatch emitted, dependency recorded
MARK    rev=1            in Realloc, ~this
FINISH  rev=1  slot=0  max=3             normalized to 3
INIT    rev=1   3 → -1   fullRebuild=0   ← StartExtension; flag wiped
FINISH  rev=1  slot=-1 max=3             no MARK in between: nothing regenerated
CHECK   rev=3  slot=-1 max=6             next compile: reads as "no dependency", skipped
```

`Add` then called `Free` through a vdata offset laid out for 3 slots against a table laid out for 6, before `main` ran.

### The fix

`StartExtension` already saves and restores `mOnDemandMethodCount` around that `Init(false)` for exactly this reason. This does the same for `mUsedSlotCount` — five lines, one file.

### Numbers

Rebuilds triggered by the slot-count change in Test1's `Virtuals.txt`:

| condition | modules rebuilt | Linux `Virtuals.txt` | Windows |
|---|---|---|---|
| upstream — `(mUsedSlotCount >= 0) && (…!= max)` | 48 | **crash** before `main` | pass |
| previous version of this PR — drop the `>= 0` guard | **1495** | pass | pass |
| **this PR** — preserve the flag across `StartExtension` | **53** | pass | pass |

The five modules added over upstream are exactly the ones whose flag was being erased: `List<int>`, `List<Object>`, `List<delegate void()>`, `List<ErrorHandlerData.ErrorHandler>`, `Event<delegate void()>` — all generic specializations, which is the population `StartExtension` operates on. Every one of upstream's 48 is still rebuilt.

Windows, for each of the three rows: `Test1/Virtuals.txt` and `SlotTest/SlotTest001.txt` run in the IDE, and `IDEHelper/Tests` 136/136, all under MSVC on Win11 — every cell passes. **Correction:** that is not because the bug is Linux-specific. Those two scripts happen not to trigger it on Windows; the minimal program in #2529 does — without this fix its second run dies with `0xC0000005` before `Main`, on Windows as well as Linux.

Also verified on Linux: `IDEHelper/Tests` 136/136, and the full Test1 IDE suite 41/41 (with the LLDB work from #2527, which the HotSwap scripts need to run at all).


bfiete:
I do not expect the IDE tests to be able to pass on Linux - GDB/LLDB will likely not be able to ever reproduce all the features of the Windows debugger, including virtual or interface dispatch.

Can you provide a test that causes a failure other than the results of a debugger string?

robknopf:
Fair on the IDE tests — I won't argue LLDB will ever match the Windows debugger feature-for-feature.

On your actual question, I went looking for a non-debugger test and should report what I found honestly: **this doesn't reproduce on Windows at all.** I built `IDEHelper` on a Win11 machine both ways and ran the two relevant scripts:

| | `Test1/Virtuals.txt` | `SlotTest/SlotTest001.txt` |
|---|---|---|
| with the fix | pass | pass |
| **without the fix** | **pass** | **pass** |

So I can't give you a test that fails on your platform, and I'd rather say that than dress up a Linux-only repro as something more.

What I'd argue instead is that the guard is unsound regardless of whether it currently bites:

```cpp
if ((module->mUsedSlotCount >= 0) && (module->mUsedSlotCount != mCompiler->mMaxInterfaceSlots))
```

`mUsedSlotCount == -1` means *"never marked"*, not *"independent of the slot count"*. `HadSlotCountDependency()` is only reached where codegen notices a dependency, so `-1` is absence of evidence rather than evidence of absence — a module holding stale vdata offsets that was never marked gets skipped. Which modules end up marked varies per build, which is presumably why Windows escapes and Linux doesn't.

The Linux symptom was a binary that crashed before reaching `main` — corlib's `List<T>.Add` calling its virtual `Free` through vdata offset `0x78` when the rebuilt vdata had moved it to `0xa0`. Only after adding interfaces to a type *in a running IDE session*; a clean build or BeefBuild (fresh process per compile) never showed it.

For what it's worth on the "is Linux worth it" question, I measured the Test1 suite there:

| | result |
|---|---|
| clean checkout | **24 / 41** |
| `+ AllowHotSwapping = true` for `Linux64` | **41 / 41** |

All 17 of those failures are one default — `Workspace.bf:1250` hard-codes `AllowHotSwapping` to `(platformType == .Windows) && !isRelease`, so every `HotSwap_*` script reports "Hot Compilation is not enabled" and never gets to exercise anything.

Two honest caveats on that number: it needs the LLDB work, which I haven't submitted yet, so it isn't something this PR delivers on its own. And `HotSwap_ExtensionOverride` hung once at the link step during a run, then passed three times in a row after — so it's 41/41 with one flake I haven't chased down.

If a test that only fails on Linux is still useful, I can add one that asserts on the program's own output instead of the debugger: compile, toggle the interfaces on, recompile, run the binary, check a sentinel it writes. That fails on the crash with no debugger involved. Equally happy to drop this PR if you'd rather not carry a fix for something you can't currently observe.

robknopf:
For context on the Linux numbers above: I have now opened #2527 with the LLDB work.

That makes this PR concrete rather than theoretical — #2527 gets `IDE/Tests/Test1` to 40/41 on Linux on its own, and the one remaining failure is `Virtuals.txt`, which needs the fix here. Together they are 41/41.

So if #2527 is of interest, this is the piece that finishes it. If it is not, this stays what I described above: a guard that treats "never marked" as "independent of the slot count", with no way for me to make it fail on Windows.

bfiete:
Yes, a Linux-only failure is useful. I do not believe that this is a proper fix, as I believe it over-rebuilds. More investigation is required.

robknopf:
You are 100% right to push back on that.  Running off my local fork was 'good enough for me', but it was wasteful and really deserved deeper investigation.  Honestly, I couldn't find what which module was causing the issue, so I burned some of those precious Fable tokens to help me out. :) You were right, and I've replaced the PR. The `>= 0` guard was heavy handed and it rebuilt 1495 modules to catch one.

Tracing `mUsedSlotCount` on the crashing module showed where the dependency actually goes missing: `StartExtension()` re-enters a finished module via `Init(false)`, and `Init` resets the flag unconditionally while the extension keeps the code that recorded it. The module then reads as slot-independent on the next compile and keeps its old offsets.

`StartExtension` already saves and restores `mOnDemandMethodCount` around that same call; the fix does the same for `mUsedSlotCount`. Five lines in `BfModule.cpp`, and `UpdateRevisedTypes` is untouched.

| condition | modules rebuilt | Linux | Windows |
|---|---|---|---|
| upstream | 48 | crash | pass |
| previous PR (drop the guard) | 1495 | pass | pass |
| this PR (preserve across extension) | 53 | pass | pass |

Windows column is `Virtuals`, `SlotTest001` and `IDEHelper/Tests` under MSVC for every row; Linux column is `Virtuals.txt`.

Full details, the flag trace and the five affected modules are in the rewritten description.

robknopf:
The Linux-only test is up as #2529. It asserts on a file the program writes when it reaches `Main`, not on anything read through the debugger, and the second run isn't launched under the debugger at all. Without the fix here it fails on Linux with `Failed to read file '…/slot_sentinel.txt'`; with it, it passes on both platforms.

bfiete:
Thanks for the test. I have fix commited at ea52d1c64b19553a51232f5ea9953304513b53c6

The rebuilding is more constrained. I did not do the actual Linux test. It would be pretty funny if that still didn't work.

Claude says to tell your Claude this:

Thanks for tracking this down. The symptom analysis is right, and #2521 does fix the crash, but we're going with a narrower fix because the mechanism turned out to be different. The slot dependency isn't going unrecorded: `HadSlotCountDependency()` does run on the affected module (`System_GC` in the #2529 repro), but `BfModule::Init` later runs again on that same module while its generated code is retained and resets `mUsedSlotCount` to `-1`, and `Finish` only refreshed the count when it wasn't `-1`. So the mark was erased rather than missing. Treating `-1` as "rebuild" works, but `-1` is also the state of every module with no slot dependency at all, so #2521 rebuilds nearly everything on a slot-count change (roughly 1,600 modules in the two-file test project, where exactly one needed it) and does so again on every later change. Our fix keeps a separate dependency flag that survives `Init`, has `Finish` record the slot count the module was actually built with, and leaves the rebuild condition narrow. On the #2529 test the second compile takes about 0.85s versus 2.15s with #2521 (0.73s on broken master), with the same result. One note on #2529: the bug does reproduce on Windows, as an access violation through a null slot in `System.GC.__BfStaticDtor`; the test catches it via `Execute`'s exit code rather than the sentinel, since the crash lands after `Main`, and a stale `build` directory hides it

bfiete:
You know, I do believe I've seen this occur occasionally. Every once in a while something random just starts crashing that was nowhere near what I was modifying (fixed by doing a Clean), and in reprospect I think it's a virtual call in every case.

robknopf:
My Claude agrees that your Claude is right:

Confirmed on Linux — `ea52d1c6` passes the #2529 test there, both compiles and both runs clean. So no, it's not funny.

Your mechanism is better than mine. I preserved `mUsedSlotCount` itself across `Init`, which conflates "has a slot dependency" with "was built at count N"; a separate flag keeps the count honest, and mine would have kept over-rebuilding on every later change. On the numbers — I measured 53 modules rather than ~1600, but that was Test1's Virtuals scenario, not the #2529 project, and I quoted it as though it generalised.

One thing worth knowing about that test, though: **it only exercises half of your commit.** Reverting only the `BfContext.cpp` condition, leaving `EmitDynamicCastCheck` alone, still passes. The `HadSlotCountDependency()` you added on the dyn-cast path is what this repro actually needs — with it, the original `mUsedSlotCount >= 0` condition suffices. So `mHadSlotCountDependency` surviving `Init` isn't covered by #2529. I haven't managed to isolate a case that hits it.

I also couldn't reproduce the stale `build` directory hiding the failure on Linux — reverted, it failed identically with a stale directory and a cleared one. Might be Windows-specific.

> Every once in a while something random just starts crashing that was nowhere near what I was modifying (fixed by doing a Clean)

That fits — stale vdata offsets in a module you never touched, cleared by a rebuild.

PR #2523, opened 2026-09-20
wasm: configurable Emscripten, an optimized runtime, and separate debug runtimes

robknopf:
Three fixes to the wasm32 build path, found while shipping a wasm target from Linux.

**1. Link with a user-configured Emscripten.** A wasm32 link required `EmsdkDep2_Done.txt` in `<install>/../wasm` for *any* configured Emscripten path, but only `fetch_wasm.bat` writes that marker, when it unpacks the bundled emsdk into `wasm/emsdk`. An Emscripten installed separately and set as `EmscriptenPath` could therefore never link — BeefBuild reported "Emscripten path not configured" although the path was set. The marker is now required only when `EmscriptenPath` *is* the bundled emsdk; any other path is validated by its `emcc`, with an error naming the missing compiler. `install.sh` also installs the wasm runtime archives (`Beef*RT32_wasm*.a`), since BeefBuild links against them from its bin directory.

**2. Build the runtime optimized.** `build_wasm.sh`/`.bat` compiled the runtime with `-g` and no optimization level, so every wasm32 program — Release included — linked an unoptimized runtime. They now build `-O2` by default (`release`); `debug` builds `-O0 -g` for debugging the runtime itself; `setup` is unchanged; anything else prints usage and fails. `Beef042RT32_wasm.a` goes from 1.87 MB to 562 KB.

**3. Separate debug runtimes, chosen by Beef Lib Type.** Every wasm32 program linked the one `Beef<ver>RT32_wasm[_pthread].a`, so a debug runtime meant rebuilding that archive for every program. The debug runtimes are now their own archives (`..._d.a`), linked when the project's Beef Lib Type is `DynamicDebug` — the same setting that selects `libBeefRT_d.a` on Linux and macOS. `BuildContext.GetWasmRtLibName` builds the name for both `$(LinkFlags)` and the pre-link existence check; that check previously only looked for the plain archive, so a threaded project with a missing runtime failed inside `emcc` instead of with a clear message.

### Testing

Checked on Linux with BeefBuild: the default links the optimized runtime, `DynamicDebug` links `_d`, a missing runtime is reported by name, and an emsdk path with no marker links. `build_wasm.sh` was run for `release`, `debug` and a bad argument. `build_wasm.bat` follows the same logic but I have no Windows machine to run it on — worth a look from someone who does.

https://claude.ai/code/session_01DHodcr5rq5c3MTNoUVdaWY


bfiete:
I believe this looks correct.

bfiete:
Ugh that means I need to do another wasm forced release though now I guess.

bfiete:
We're updated to EmsdkDep3 now...

PR #2524, opened 2026-09-20
SDL: render transparent windows with their own shared GL context

robknopf:
On Linux/SDL, `DestAlpha` windows (menus, tooltips, autocomplete) are created with `SDL_WINDOW_TRANSPARENT` and get an ARGB visual. The single GL context was created for the main window's opaque visual, and NVIDIA's GLX refuses to make it current on an ARGB window (`BadMatch`). Those windows never presented, and the failed `MakeCurrent` left the context in a bad state, so later windows (dialogs) came up black too.

Transparent windows now get a second context that shares the first one's objects. Each `GLRenderWindow` keeps its context, VAOs are per context, and GL state is reset to the frame-start baseline when switching contexts.

Also fixes menu and autocomplete placement on monitors not at x=0: the right-edge checks compared screen X against the workspace width without adding the workspace's X offset, so a menu on a second monitor always flipped to the left.

### Testing

Linux, NVIDIA proprietary driver, multi-monitor: menus, tooltips, autocomplete popups and dialogs all render correctly, including on a second monitor placed to the right of the primary. I don't have an AMD/Intel or a macOS machine to check those on.

https://claude.ai/code/session_01DHodcr5rq5c3MTNoUVdaWY


bfiete:
Are we 100% sure this Menu.bf changes doesn't break multimon setups on Windows?

robknopf:
I originally did testing to make sure I didn't break Windows via Wine (Proton), but to be sure, I fired up my old Win11 laptop and built it there to make sure it was on real hardware.  Full disclosure:  I asked Claude to give a summary and provide a screenshot of A/B.  I didn't see any restrictions regarding LLM usage, let me know if this is out of bounds.

**Why the primary monitor can't be affected**

`GetWorkspaceRectFrom` returns `MONITORINFO.rcWork` in *virtual-desktop* coordinates (`WinBFApp.cpp`), so `workspaceX` is the work area's left edge rather than 0. On the primary monitor `workspaceX == 0`, which makes the old and new expressions arithmetically identical — single-monitor and primary-monitor behaviour is bit-for-bit unchanged.

It also brings X in line with what the Y axis in that same function already does: `maxY = workspaceY + workspaceHeight` two lines up, and `wsY + wsHeight` in `Init`. X was the only spot missing its origin term.

**What happens on a secondary monitor**

Test setup: `DISPLAY1` work area `{X=-1920, W=1920}` (positioned left of primary), `DISPLAY4` primary `{X=0, W=1920}`. IDE on DISPLAY1, context menu opened 150px from that monitor's right edge, menu width 224:

- old: `-150 + 224 = 74 > 1920`? → false → no flip
- new: `74 > -1920 + 1920 = 0`? → true → flips left

Measured menu positions, identical window and cursor, screen coordinates:

| | menu x-range | result |
|---|---|---|
| before | `0 … +224` | **entirely on the other monitor** |
| after | `-372 … -145` | stays on DISPLAY1 |
menu_ab

**Why it jumps monitors rather than merely overhanging**

Without the flip the popup is created straddling the seam at `x=0`, so `EnumDisplayMonitors` invokes `ClipToMonitor` (`WinBFApp.cpp`) for *both* monitors against one shared rect:

1. DISPLAY1 (work `-1920…0`): `-152 + 224 >= 0` → `mX = BF_MAX(-1920, 0-224)` → `-224`
2. DISPLAY4 (work `0…1920`): `-224 < 0` → `mX = 0`

The second monitor's clamp overwrites the first, landing the menu fully on DISPLAY4. That reproduces the measured `0…224` exactly. With the fix the popup never intersects DISPLAY4, so the clamp is a no-op and `-372` survives untouched.

On a side note: the cumulative, enumeration-order-dependent behavior in `ClipToMonitor` looks like it might eventually be problematic. This change just stops popups from tripping it.  Let me know if you want that as a separate PR.

bfiete:
Thanks for checking. I'm not double-checking the rest of the implemnetation, as the Linux IDE itself is fairly far from "production-ready" and in need of contributions.

PR #2526, opened 2026-09-21
Take the comptime allocation before subtracting the memory base

jayrulez:
CreateAttribute and the params span path computed a CE address as 'CeMalloc(...) - mMemory.mVals' in one expression. The allocation can grow mMemory, which moves mVals, and the operands of '-' are unsequenced: GCC and Clang evaluate the call first, MSVC reads the base first, so on Windows an attribute whose allocation was the one to grow the buffer got the new pointer minus the old base. Its constructor then failed the interpreter's bounds check as an access violation, and the type apply that followed wrote through the same address.

Only the allocation that crosses the buffer's capacity is affected. Two statements now, at both sites.

The test runs a comptime job large enough that the context's memory is not carried over, then constructs an attribute in the context that follows, whose allocation is then the one that grows the buffer.


bfiete:
Yah, I forget those C++ sequencing rules sometimes, or rather "lack of rules".

PR #2527, opened 2026-09-21
LLDB: Linux debugger support, including hot compilation

robknopf:
Builds out the LLDB debugger backend so the IDE is usable on Linux, including hot compilation.

This is large, so it is arranged as 23 reviewable commits rather than one drop. It is almost entirely confined to the LLDB backend:

```
 BeefRT/dbg/DbgInternal.cpp |  162 +-
 IDEHelper/DebugManager.cpp |    5 +
 IDEHelper/LLDBDebugger.cpp | 5848 ++++++++++++++++++++++++++++++++++++++++++--
 IDEHelper/LLDBDebugger.h   |  219 +-
```

plus a three-line test workspace config (see below). Nothing in the Windows path is touched.

### Hot compile / hot swap

The bulk of the work. Hot compiling on Linux, target output forwarding, runtime type tables updated in place the way WinDebugger does it, debug info for hot-loaded code and safe stepping into it, step traps so small replaced methods can still be stepped into, per-compile breakpoint binding, lambda capture checks, thread-local variables, hot-swap memory reuse, and support for programs built as shared libraries.

### Evaluator

Calling Beef methods, properties and indexers, Beef-aware member paths, String display, format specifiers, static fields, and autocomplete covering default arguments, enum arguments and hot-compiled methods.

### Stepping and breakpoints

Step filters, step-out that finishes the caller's line, step-out landing in a scope's cleanup code, no stopping or breakpoint binding on non-statement rows, disassembly breakpoints, restarting after a stop, a modules panel, set-next-statement, and logging breakpoints.

### Defaults

Uses LLDB where there is no WinDebugger available.

### Testing

`IDE/Tests/Test1`, run headless under Xvfb on Ubuntu with LLVM/LLDB 22:

| | result |
|---|---|
| this branch alone | **40 / 41** |
| this branch + #2521 | **41 / 41** |

The one difference is `Virtuals.txt`, which needs the interface-slot rebuild fix in #2521 — unrelated to the debugger, but it is what that script trips over on Linux.

The last commit adds `[Configs.Debug.Linux64] AllowHotSwapping = true` to the Test1 workspace. `AllowHotSwapping` defaults to `(platformType == .Windows) && !isRelease` in `Workspace.bf`, so without it every `HotSwap_*` script fails with "Hot Compilation is not enabled" and the suite reads 24/41. The `Win64`/`Win32` configs are already listed there; this adds the Linux one and leaves the product default alone. Happy to flip that default for Linux in a separate PR if you would rather have it on generally now that hot swap works there.

Two things I would rather state than have you find:

- `HotSwap_ExtensionOverride` hung once at the link step (`Executing (IDETest): /usr/bin/c++`) during a full run, then passed three times in a row afterwards. One flake in roughly five executions; I have not tracked it down.
- All of this was measured under Xvfb rather than a real display.

I do not claim parity with the Windows debugger. The aim is a Linux IDE that is usable day to day, and I am happy to split any part of this out or rework it if the shape is wrong.


bfiete:
Well, I have previously stated that it would be impossible to create hot compiling on Linux and... I guess here's at least a partially-working form of that? That's pretty crazy.

At any rate, I think it's unlikely to *break* anyone's LLDB experience because it wasn't really working that great to begin with. Hm.

PR #2529, opened 2026-09-21
Add a debugger-free test for the interface slot rebuild

robknopf:
The test promised on #2521. Now that `ea52d1c6` is in, this is the regression test for it.

`IDE/Tests/SlotRebuild` mirrors Test1's `Virtuals.bf`: compile, run, uncomment four interfaces on `ClassB` and their uses, compile, run. The program asserts the result of its virtual calls; a failed assert or a crash exits non-zero, and `Execute` fails the script on a non-zero exit. Nothing is read through the debugger, and the runs are not launched under it.

```
Compile()
Execute("$(TargetPath)")

ToggleCommentAt("ClassA_IFaces")
ToggleCommentAt("Test_IFaces")
Compile()
Execute("$(TargetPath)")
```

Rebased now that the `ScriptManager` changes this needed are upstream as `fe0d8d03`, so this is just the test project.

### Verified against `ea52d1c6`

On Linux, with a cleared `build` directory each time:

| compiler | result |
|---|---|
| `ea52d1c6` | **pass** |
| `ea52d1c6` fully reverted | **fail** — `exited with code '139'` |

So it catches the bug on current master and passes with the fix. On Windows without the fix it fails as `0xC0000005`.

### One caveat on what this covers

`ea52d1c6` has two parts, and this test only exercises one of them. Reverting *only* the `BfContext.cpp` condition — leaving the `EmitDynamicCastCheck` changes in place — still **passes**. The `HadSlotCountDependency()` call added on the dyn-cast path is what this repro needs; with it, the original `mUsedSlotCount >= 0` condition is sufficient.

The `mHadSlotCountDependency` flag surviving `Init` therefore isn't covered here. It addresses a module whose mark is erased while its code is retained, which this program doesn't appear to hit. Worth knowing if you want that path tested too — I haven't managed to build a case that isolates it.

I also couldn't reproduce a stale `build` directory masking the failure on Linux: reverted, it failed identically with a stale directory and a cleared one.


bfiete:
Wouldn't it have been easier to just make a program that did a Test.Assert on what would have been an obviously correct value instead of doing all this file stuff?

bfiete:
Good news is that it just crashed the app on Windows so that's even better than files

robknopf:
You were right on both counts, and the second one was more useful than expected.

**Asserting instead of files:** done — the program now `Runtime.Assert`s its virtual-call results and the script just runs it twice. The reason I'd reached for a file is that on Linux it *didn't* work: `BfpSpawn_WaitFor` reports a signal-killed child as exit code 0 (`WEXITSTATUS` on a `WIFSIGNALED` status), so the crash never surfaced to `Execute`. That's a real runtime bug for anything spawning children on POSIX; fixed in #2530 (`128 + signal`, as a shell does). With it, the Linux run fails with `exited with code '139'`.

**"It just crashed the app on Windows":** that's the test program crashing without the fix — `0xC0000005` before `Main`. Which means the bug reproduces on Windows after all; `Virtuals.txt` and `SlotTest001` simply never triggered it there. I've corrected #2521's description, which claimed otherwise.

| compiler | Linux | Windows |
|---|---|---|
| with #2521 | pass | pass |
| without | fails (`139`) | fails (`0xC0000005`) |

bfiete:
we don't need this anymore

